PR #961, opened 2015-01-18
Duplicate types from CSharpGenerator

Layoric:
Bug causing duplicate types due to not adding processed type to existingOps hashset.

### DIFF
--- a/src/ServiceStack/NativeTypes/CSharp/CSharpGenerator.cs
+++ b/src/ServiceStack/NativeTypes/CSharp/CSharpGenerator.cs
@@ -146,6 +146,7 @@ namespace ServiceStack.NativeTypes.CSharp
                 {
                     lastNS = AppendType(ref sb, type, lastNS, allTypes, 
                         new CreateTypeOptions { IsType = true });
+                    existingOps.Add(fullTypeName);
                 }
             }
 


mythz:
ok cool, thx Darren

mythz:
Hey Darren, IMO it should be added to `types` since the Type is not an operation DTO, but do you have a DTO example where it prevents duplicate types?

mythz:
Actually we'd need to rename `expstingOps` to `existingTypes` so it's used for both, I'll make the change.

mythz:
Can you let me know the DTO's used that caused duplicate types and this fix prevents? I haven't been able to repro it and I want to add it to the existing test DTO's to catch regressions.

Layoric:
Object that is a property of a request object multiple times. See service model below.

```csharp
/// <summary>
/// Duplicate
/// </summary>
public class TypeB
{
    public string Foo { get; set; }
}

public class TypeA
{
    public List<TypeB> Bar { get; set; } 
}

/// <summary>
/// Request object 1
/// </summary>
public class Create : IReturn<CreateResponse>
{
    public TypeA Test { get; set; }
}

public class CreateResponse
{
    public TypeA Test { get; set; }
}

/// <summary>
/// Request object 2
/// </summary>
public class Update : IReturn<UpdateResponse>
{
    public TypeA Test { get; set; }
}

public class UpdateResponse
{
    public TypeA Test { get; set; }
}
```

Services
``` csharp
public class MyServices : Service
{
    public object Get(Create request)
    {
        return new CreateResponse();
    }

    public object Get(Update request)
    {
        return new UpdateResponse();
    }
}
```

Results with

``` csharp
/* Options:
Date: 2015-01-19 07:25:10
Version: 1
BaseUrl: http://localhost:46039

//MakePartial: True
//MakeVirtual: True
//MakeDataContractsExtensible: False
//AddReturnMarker: True
//AddDescriptionAsComments: True
//AddDataContractAttributes: False
//AddIndexesToDataMembers: False
//AddResponseStatus: False
//AddImplicitVersion: 
//InitializeCollections: True
//AddDefaultXmlNamespace: http://schemas.servicestack.net/types
*/

using System;
using System.Collections;
using System.Collections.Generic;
using System.Runtime.Serialization;
using ServiceStack;
using ServiceStack.DataAnnotations;
using WebApplication7.ServiceModel.Types;


namespace WebApplication7.ServiceModel.Types
{

    public partial class Create
        : IReturn<CreateResponse>
    {
        public virtual TypeA Test { get; set; }
    }

    public partial class CreateResponse
    {
        public virtual TypeA Test { get; set; }
    }

    public partial class TypeA
    {
        public TypeA()
        {
            Bar = new List<TypeB>{};
        }

        public virtual List<TypeB> Bar { get; set; }
    }

    public partial class TypeB
    {
        public virtual string Foo { get; set; }
    }

    public partial class TypeB
    {
        public virtual string Foo { get; set; }
    }

    public partial class Update
        : IReturn<UpdateResponse>
    {
        public virtual TypeA Test { get; set; }
    }

    public partial class UpdateResponse
    {
        public virtual TypeA Test { get; set; }
    }
}
```

PR #962, opened 2015-01-24
Use lookup table for GetSubscription and GetSubscriptionInfosByUserId

Nness:
MemoryServerEvents use lookup table for GetSubscription and
GetSubscriptionInfosByUserId instead of loop through all entities.

### DIFF
--- a/src/ServiceStack/ServerEventsFeature.cs
+++ b/src/ServiceStack/ServerEventsFeature.cs
@@ -554,13 +554,13 @@ namespace ServiceStack
         public IEventSubscription GetSubscription(string id)
         {
             if (id == null) return null;
-            foreach (var subs in Subcriptions.Values)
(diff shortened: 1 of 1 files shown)


mythz:
Ahh yes, great catch!

thx for the PR :+1:

Nness:
There is another one that I didn't modify. GetSubscriptionsDetails(params string[] channels)
Since latest support multiple channels for same subscription, to get that list with dictionary will most likely not simple list add, but use hash table. I'm not sure if that's best approach to resolve that.

So just point out here if you want to take look. :D 

Have been work closely with ServerSentEvent recently. Glade that can help you out.

mythz:
Yep no worries, should be sorted in this commit: https://github.com/ServiceStack/ServiceStack/commit/c4606f6a700b74e535547d097a9d588a1325faf5

PR #963, opened 2015-01-31
Microsoft Account OAuth 2 Provider added.

ivanfioravanti:
Note:
- Gender is always null even if defined online in Microsoft Account site 
- http://localhost:11001 cannot be used as RedirectURL, for testing you should use something like http://devmachine.local:11001/


mythz:
Awesome! thx Ivan :+1:

PR #964, opened 2015-02-05
Add Instagram OAuth2 Provider.

ricardobrandao:
We need to include System.Runtime.Serialization to the project in order to increase the default max depth quota.


mythz:
Awesome! thx for PR :+1:

PR #965, opened 2015-02-05
Allow multiple newlines to be published on ServerEvents and silently consumed by the C# ServerEventsClient.

jklemmack:
In some situations, response buffering is enabled on a server, causing ServerEvent messages to not be fully transmitted to the client.  One workaround is to remove dynamic compression, at a potential cost to all other API calls on the site.  The other option, [outlined by mythz here](http://stackoverflow.com/questions/25960723/servicestack-sever-sent-events/25983774#25983774), is to write an amount of bogus `\n` chars to the output stream.  This alternative locks up the `ServiceStack.ServerEventsClient` when attempting to digest the `\n` chars.

I'm not entirely happy with the logic, but it passes the unit tests.

    OnPublish = (res, msg) =>
    {
        res.OutputStream.Write("\n\n\n\n\n\n\n\n\n\n");
    }

### DIFF
--- a/src/ServiceStack.Client/ServerEventsClient.cs
+++ b/src/ServiceStack.Client/ServerEventsClient.cs
@@ -389,14 +389,16 @@ namespace ServiceStack
                     {
                         if (text == "\n")
                         {
-                            ProcessEventMessage(currentMsg);
+                            if (currentMsg != null) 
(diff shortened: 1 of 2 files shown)


mythz:
great, thx Johann :+1:

PR #967, opened 2015-02-12
Add failing test for Db.Dispose() in a razor partial.

jeffgabhart:
Add a breakpoint at [line 515](https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack.Razor/ViewPageBase.cs#L515) in `ViewPageBase.cs`
Browse to `localhost:59115/catchall`
`.Dispose()` is called for CatchAll and SimplyLayout, but not for CatchAllPartial
This causes leaked db connections which, over time, will end up exhausting the pool and crashing the site.

### DIFF
--- a/tests/ServiceStack.RazorHostTests/CatchAll.cshtml
+++ b/tests/ServiceStack.RazorHostTests/CatchAll.cshtml
@@ -20,3 +20,5 @@
         <li>@person.FirstName - @person.LastName (@person.Age)</li>
     }
 </ul>
+
+@Html.Partial("_CatchAllPartial")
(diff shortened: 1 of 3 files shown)


mythz:
ok thx, will check it out

mythz:
Well it's not exactly a failing test, but partials should now be disposed with this commit:
https://github.com/ServiceStack/ServiceStack/commit/fa416727087411e030751c613116976388fcdbe2

This change is now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

jeffgabhart:
Yeah, that's true. Thanks

PR #972, opened 2015-02-25
Handle invalid type used with appHost.Routes.Add<T>

Layoric:
Handle invalid type used with appHost.Routes.Add<T>. Currently throws a null ref error if type is not used in service.

### DIFF
--- a/src/ServiceStack/Support/WebHost/FilterAttributeCache.cs
+++ b/src/ServiceStack/Support/WebHost/FilterAttributeCache.cs
@@ -43,6 +43,10 @@ namespace ServiceStack.Support.WebHost
             var attributes = requestDtoType.AllAttributes<IHasRequestFilter>().ToList();
 
             var serviceType = HostContext.Metadata.GetServiceTypeByRequest(requestDtoType);
+            if (serviceType == null)
+            {
+                throw new Exception("No service using type '" + requestDtoType.FullName + "'.");
+            }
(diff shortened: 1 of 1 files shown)


mythz:
Hey Darren, this is inconsistent to how we're [handling Response Attributes](https://github.com/Layoric/ServiceStack/blob/0abec583ff5877843a2b75e01ff59391b52e0686/src/ServiceStack/Support/WebHost/FilterAttributeCache.cs#L76) so I've changed it so they retain the same behavior in this commit: https://github.com/ServiceStack/ServiceStack/commit/c9053561064cf4488a43621aa60c4fe41e7c5d38

I've ended up moved the Exception closer at registration where it's happening and included a test so we can assert what the exception is for.

mythz:
Actually throwing an Exception is not going to work since they can still register a route in `AppHost.Configure()` for a Service that doesn't exist yet, i.e for a dynamically added Service registered in a plugin like the Registration Service.

How were you adding `appHost.Routes.Add<T>()` to get this exception?

Layoric:
Creating a plugin and registering a route in the "Register(IAppHost appHost)" function of the IPlugin interface. 

Eg,
`appHost.Routes.Add<DiscourseSsoInitialPayload>("/sso/discourse", ApplyTo.Get);`
where `DiscourseSsoInitialPayload` was not used as the request DTO of any service.

PR #974, opened 2015-03-04
Added SS.Caching.AwsDynamoDbV2 and update deps

KevinHoward:
Added new ServiceStack.Caching.AwsDynamoDbV2 project that uses version 2.0+ of the 3rd party dependency and locked the current ServiceStack.Caching.AwsDynamoDb library to use only NuGet packages version prior to 2.0. 

Updated several NuGet packages in other projects to the most recent version of their 3rd party dependencies.


mythz:
Awesome thx Kevin! much appreciated :+1:

PR #975, opened 2015-03-06
Added a FourSquare OAuth2 Provider

KevinHoward:
Added a FourSquare OAuth2 Provider and made a small change to the Microsoft Live OAuth2 provider to simplify code. I haven't fully tested this provider and I'm not sure if I handled the population of the auth tokens correctly. There may be a better way to do this - I would love feedback.

### DIFF
--- a/src/ServiceStack.Authentication.OAuth2/FourSquareOAuth2Provider.cs
+++ b/src/ServiceStack.Authentication.OAuth2/FourSquareOAuth2Provider.cs
@@ -1,0 +1,198 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using DotNetOpenAuth.Messaging;
+using DotNetOpenAuth.OAuth2;
+using ServiceStack.Auth;
+using ServiceStack.Configuration;
+using ServiceStack.Text;
+
+namespace ServiceStack.Authentication.OAuth2
+{
+    /// <summary>
+    /// More info at: https://developer.foursquare.com/overview/auth.html
+    /// </summary>
+    public class FourSquareOAuth2Provider : OAuth2Provider
+    {
+        public const string Name = "FourSquare";
+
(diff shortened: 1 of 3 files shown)


mythz:
Awesome thx for this Kevin, I'll test this out when I get some time and let you know of any issues. 

BTW I just use [ServiceStack.AuthWeb.Tests](https://github.com/ServiceStack/ServiceStack/tree/master/tests/ServiceStack.AuthWeb.Tests) to test different OAuth providers. There's also the [MVC OAuth Test Project](https://github.com/ServiceStackApps/LiveDemos#mvc) which you can plug in and test existing OAuth providers. I've still got it on my plate to create a new LiveDemo with all AuthProviders that others can try out and add in their own OAuth keys.

mythz:
Hey Kevin, I've made some tweaks to the provider in this commit: https://github.com/ServiceStack/ServiceStack/commit/718fa7ae2b87961082d033b5f85ea0af3d201d45

I've also added it to the AuthWeb test project and it looks like it works, the one thing that didn't work is the profile url which for my profile is `https://irs1.4sqi.net/img/user/121914152-P1X4QCFQFPGJ2D41` but for some reason FourSquare is returning a `500 Internal Server Error` which is strange because this is also the url in the raw JSON so I'm not sure if this is a temporary issue with FourSquare or there's something we're missing to create the profileUrl.

KevinHoward:
I'll take a look at it this evening. Thanks for the changes.
One quick question I have on the profile issue, is the URL supposed to be for your profile, or for the profile image? The user id and dash before the file name don't appear to be right. I may have done the setting of the ProfileURL incorrectly.

One other quick point. Foursquare provides a tie to the user's Facebook and Twitter if provided. I set these in the User Session, is this correct or should I have done this another way?

mythz:
The profile url is for the image profile url. Adding it to the authInfo dictionary is fine, it ends up being added to both the session and UserAuthDetail DB table.

PR #976, opened 2015-03-07
Fixed FourSquare OAuth2 Profile Image URL

KevinHoward:
Added missing profile image dimensions to the URL. Defaulted the dimensions to 64x64, but the developer can set them using oauth.FourSquare.ProfileImageWidth and oauth.FourSquare.ProfileImageHeight appsettings.

### DIFF
--- a/src/ServiceStack.Authentication.OAuth2/FourSquareOAuth2Provider.cs
+++ b/src/ServiceStack.Authentication.OAuth2/FourSquareOAuth2Provider.cs
@@ -23,6 +23,10 @@ namespace ServiceStack.Authentication.OAuth2
         //   the v (i.e. Version) parameter is required for all API requests. 
         public DateTime Version { get; set; }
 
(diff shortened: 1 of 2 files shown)


mythz:
Ahh brilliant thx Kevin, can confirm that my profile img works now: https://irs1.4sqi.net/img/user/64x64/121914152-P1X4QCFQFPGJ2D41

KevinHoward:
Demis, 

I see another unrelated issue when I was reviewing this one. 

My birthday appears to be one day off on the AuthWebTests site, but when I step through the code, the DateTime is the correct result when setting the birthday. Probably and issue with the test project's output.

Also, as side note, .NET 4.6 will have built-in support for UnixTime conversion (https://www.visualstudio.com/en-us/news/vs2015-preview-vs#Net).

KevinHoward:
I can confirm that the profile img url works.

PR #978, opened 2015-04-02
MiniProfiler: Show SQL parameter names, types, and values below the command string

schmidt4brains:
@mythz This is a tweak to the profiler output that will show the parameter values used in a SQL statement.

The MiniProfiler was already capturing the bound parameter values, it just wasn't displaying them. So I added conditional block in the template that displays each parameter/value on a line below the command string.

As you can see from the included screenshot, the CSS styling is not right. It's leaving huge whitespace between each parameter.

I really don't like raising half a pull request, but my CSS skillz are just not up to par. So after struggling with CSS for an hour, I decided to raise the PR and maybe you or another SS dev can finish the styling part?

image


mythz:
ok I'll take a look, thx Doug.

mythz:
FYI, I've made subtle tweaks in this commit: https://github.com/ServiceStack/ServiceStack/commit/5fa911a907a64892040820af994e084d49c52b03

PR #979, opened 2015-04-14
Update IAuthHttpGateway.cs

jmpeltier:
At the end of April 2015 facebook will require all API calls to be on at least v2.0.    In the future it might make sense to make this configurable in web.config.    They are currently on V2.3 I think.


mythz:
Cheers :+1: Making the urls static will allow them to be overridable, changed in this commit: https://github.com/ServiceStack/ServiceStack/commit/5365fdbf8ee93200d8d5c9babdff2bca159dc99f

PR #980, opened 2015-04-16
Add "Operator" to escaped keywords for VbNet NativeTypes

Layoric:
Fix in regards to [problem posted on ServiceStack forums](https://forums.servicestack.net/t/add-servicestack-reference-created-dtos-do-not-handle-properties-using-reserved-keywords-properly-in-vb-net/690).


mythz:
Cheers, thx Darren

PR #982, opened 2015-05-07
Request Cancellation with CancellationToken

drakelord:
Support long-running request  cancellation using cancellation token.
Requests that support cancellation will include request id header. CancellationService will create CancellationSource for such requests and provide tokens for other services to monitor cancelltaion, which may be invoked by client with provided route and request id.

I added CancellationToke to IHttpRequest and it's implementations. But maybe this is not such a good idea to add something that will be used only by very few requests at such a level. I tried another approach with global request/response filter which provides the same functionality. On client-side we need to provide methods receiving CancellationToke, subscribe to cancellation request and invoke CancellationService if requested.

Service
```c#
public class TestService : Service
{
    public TestRequestResponse Any(TestRequest req)
    {
        // or just var token = RequestCancelService.GetCancellationToken(this.Request);
        var httpReq = this.Request as IHttpRequest;
        if(httpReq == null) return new TestRequestResponse();
        while(true)
        {
            httpReq.CancelToken.ThrowIfCancellationRequested();
            Thread.Sleep(100);
        }
    }    
// OR
    public TestRequestResponse Any(TestRequest req)
    {
        var token = RequestCancelService.GetCancellationToken(this.Request);
        while(true)
        {
            token.ThrowIfCancellationRequested();
            Thread.Sleep(100);
        }
    }
}
```

Client
```c#
var cs = new System.Threading.CancellationTokenSource();

client.PostAsync(new TestRequest(), cs.Token);

cs.Cancel();
```
I would like to know your opinion, maybe there is better way to implement this.

### DIFF
--- a/src/ServiceStack.Client/AsyncServiceClient.cs
+++ b/src/ServiceStack.Client/AsyncServiceClient.cs
@@ -129,7 +129,8 @@ namespace ServiceStack
             }
         }
 
-        public Task<TResponse> SendAsync<TResponse>(string httpMethod, string absoluteUrl, object request)
+        public Task<TResponse> SendAsync<TResponse>(string httpMethod, string absoluteUrl, object request, 
+            CancellationToken cancelToken = default(CancellationToken))
         {
             var tcs = new TaskCompletionSource<TResponse>();
 
@@ -153,14 +154,16 @@ namespace ServiceStack
                         tcs.SetResult(r);
                     },
                     (response, exc) => tcs.SetException(exc),
-                    wr => webRes = wr
+                    wr => webRes = wr,
+                    cancelToken
                 );
             }
             else
             {
                 SendWebRequest<TResponse>(httpMethod, absoluteUrl, request,
                     tcs.SetResult,
-                    (response, exc) => tcs.SetException(exc)
+                    (response, exc) => tcs.SetException(exc), 
+                    cancelToken: cancelToken
                 );
             }
 
@@ -173,8 +176,9 @@ namespace ServiceStack
             SendWebRequest(httpMethod, absoluteUrl, request, onSuccess, onError);
         }
 
-        private void SendWebRequest<TResponse>(string httpMethod, string absoluteUrl, object request, 
-            Action<TResponse> onSuccess, Action<object, Exception> onError, Action<WebResponse> onResponseInit = null)
+        private void SendWebRequest<TResponse>(string httpMethod, string absoluteUrl, object request,
+            Action<TResponse> onSuccess, Action<object, Exception> onError, Action<WebResponse> onResponseInit = null,
+            CancellationToken cancelToken = default(CancellationToken))
         {
             if (httpMethod == null) throw new ArgumentNullException("httpMethod");
 
@@ -189,7 +193,7 @@ namespace ServiceStack
                 }
             }
 
-            var webRequest = this.CreateHttpWebRequest(requestUri);
+            var webRequest = this.CreateHttpWebRequest(requestUri);           
 
             var requestState = new AsyncState<TResponse>(BufferSize)
             {
@@ -205,9 +209,16 @@ namespace ServiceStack
             };
             requestState.StartTimer(this.Timeout.GetValueOrDefault(DefaultTimeout));
 
+            if (cancelToken.CanBeCanceled)
+            {
+                var requestId = Guid.NewGuid();
+                webRequest.Headers.Add(CancelRequest.CancelRequestIdHeader, requestId.ToString());
+                requestState.SetCancelToken(cancelToken, () => SendCancelRequest(requestId));
+            }
+
             SendWebRequestAsync(httpMethod, request, requestState, webRequest);
         }
-
+       
         private void SendWebRequestAsync<TResponse>(string httpMethod, object request,
             AsyncState<TResponse> state, HttpWebRequest webRequest)
         {
@@ -252,6 +263,17 @@ namespace ServiceStack
                 // BeginGetRequestStream can throw if request was aborted
                 HandleResponseError(ex, state);
             }
+        }
+
+        private void SendCancelRequest(Guid requestId)
+        {
+            Console.WriteLine("Canceling...");
+            var req = new CancelRequest() { RequestId = requestId };
+            try
+            {
+                SendWebRequest<CancelRequestResponse>(HttpMethods.Put, BaseUri.CombineWith(req.ToPutUrl()), req, null, null);
+            }
+            catch (Exception ex) { Console.WriteLine("Canceling failed Error: " + ex.Message); }
         }
 
         private void RequestCallback<T>(IAsyncResult asyncResult)
--- a/src/ServiceStack.Client/AsyncState.cs
+++ b/src/ServiceStack.Client/AsyncState.cs
@@ -110,6 +110,16 @@ namespace ServiceStack
                 this.Timer = null;
             }
         }
+        CancellationTokenRegistration? cancelRegistration;
+        public void SetCancelToken(CancellationToken cancelToken, Action onCancel)
+        {
+            cancelRegistration = cancelToken.Register(onCancel);
+        }
+
+
+       
+
+
 
 #if NETFX_CORE
             public void TimedOut(ThreadPoolTimer timer)
@@ -156,6 +166,11 @@ namespace ServiceStack
             {
                 this.Timer.Dispose();
                 this.Timer = null;
+            }
+            if(cancelRegistration.HasValue)
+            {
+                cancelRegistration.Value.Dispose();
+                cancelRegistration = null;
             }
         }
     }
--- a/src/ServiceStack.Client/CancelRequestDtos.cs
+++ b/src/ServiceStack.Client/CancelRequestDtos.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace ServiceStack
+{
+    [Route("/requests/canceled")]
+    [DataContract]
+    public class CancelRequest : IReturn<CancelRequestResponse>
+    {
+        public const string CancelRequestIdHeader = "X-SS-RequestId";
+
+        [DataMember(Order = 1)]
+        public Guid RequestId { get; set; }
+    }
+
+    [DataContract]
+    public class CancelRequestResponse
+    {
+        [DataMember(Order = 1)]
+        public ResponseStatus ResponseStatus { get; set; }
+
+        public CancelRequestResponse()
+        {
+            ResponseStatus = new ResponseStatus();
+        }
+    }
+}
--- a/src/ServiceStack.Client/ServiceClientBase.cs
+++ b/src/ServiceStack.Client/ServiceClientBase.cs
@@ -15,6 +15,7 @@ using ServiceStack.Logging;
 using ServiceStack.Messaging;
 using ServiceStack.Text;
 using ServiceStack.Web;
+using System.Threading;
 
 #if !(__IOS__ || SL5)
 #endif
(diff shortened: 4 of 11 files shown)


mythz:
Hi, thx for the PR.

Whilst I liked the feature I wasn't a fan of the disruptive nature of it so I've had a go at implementing this feature with a less invasive approach in this commit: https://github.com/ServiceStack/ServiceStack/commit/676d7d02e08fe8bb3d54bb249f53c1c27763099b

I wanted to ensure that this feature was encapsulated in an optional `CancellableRequestsFeature` where it wouldn't add any overhead or complexity when this feature isn't used.

### Server Usage

To ensure this feature is localized to only where it's needed, the dictionary of tagged requests are cleaned up after use and the API clearly identifies the scope of where the request can be cancelled, I've taken a disposable pattern for usage with a Service, e.g:

```csharp
public object Any(TestCancelRequest req)
{
    using (var cancellableRequest = base.Request.CreateCancellableRequest())
    {
        while (true)
        {
            cancellableRequest.Token.ThrowIfCancellationRequested();
            Thread.Sleep(100);
        }
    }
}
```

### Client Usage

For the client usage I didn't want to pollute each async API with an additional token and it also wasn't clear that the client Token was cancelling the server operation which I preferred was clearer as to what's actually happening. I also wanted to avoid coupling and baking the implementation of this feature in the client, so I'm using a Request Filter instead.

To make use of this feature the client needs to specify a Tag for this request (inspired from your PR) which can be any string. We can use a RequestFilter to populate the Request Header with an additional `X-Tag` HTTP Header, i.e:

```csharp
var tag = Guid.NewGuid().ToString();
var client = new JsonServiceClient(Config.AbsoluteBaseUri) {
    RequestFilter = req => req.Headers[HttpHeaders.XTag] = tag
};
```

> The generic `X-Tag` Header could also be re-used in other features that require tagging and operating on requests.

To cancel this request we can just call it like a normal ServiceStack Service, providing the tag above, e.g: 

```csharp
var response = client.Post(new CancelRequest { Tag = tag });
```

As it's just a normal Service call, this can be called with the same or new client instance in the same or bg thread to cancel both sync and async requests.

This `CancellableRequestsFeature` plugin is available from v4.0.41 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet). 

Please let me know if you find any issues with it.

Thanks!

drakelord:
Thx for adding this feature! Using disposable directly in consuming service indeed is simplest, yet smartest solution.

malat:
Hi guys,

Great feature BTW 

How this will work in a load balanced environment ?

Thanks

drakelord:
Well, it won't:) But you can keep track of Tag:ExecutingServer in shared cache (redis, memcached) and redirect cancellation request with new Post/Put or something else(ServerEvents, zeromq, nanomsg...). In my opinion, this feature shouldn't implement all this, but may provide hooks for custom actions to override CancellableRequestService behavior.
e.g.
```c#
public class CancellableRequestsFeature : IPlugin
{
	public Action<string> OnRegisterCancellableRequest { get; set; }

	public Action<string> OnUnregisterCancellableRequest { get; set; }

	public Func<CancelRequest, CancelRequestResponse> RedirectCancelation { get; set; }

	internal void UnregisterCancellableRequest(string requestTag)
	{
		ICancellableRequest existing;
		if (RequestsMap.TryRemove(requestTag, out existing))
		{
			existing.Dispose();
			if (OnUnregisterCancellableRequest != null)
				OnUnregisterCancellableRequest(requestTag);
		}
	}
}

public object Any(CancelRequest request)
{
	if (request.Tag.IsNullOrEmpty())
		throw new ArgumentNullException("Tag");

	using (var cancallableReq = base.Request.GetCancellableRequest(request.Tag))
	{
		if(cancallableReq == null)
		{
			var feature = HostContext.GetPlugin<CancellableRequestsFeature>();
			if(feature.RedirectCancelation != null)
				return feature.RedirectCancelation(request);
		}

		if (cancallableReq == null)
			throw HttpError.NotFound("Request with Tag '{0}' does not exist".Fmt(request.Tag));

		cancallableReq.TokenSource.Cancel();

		return new CancelRequestResponse {
			Tag = request.Tag,
			Elapsed = cancallableReq.Elapsed,
		};
	}
}

public CancellableRequest(CancellableRequestsFeature feature, IRequest req, string tag)
{
	this.TokenSource = new CancellationTokenSource();
	this.Token = this.TokenSource.Token;
	this.feature = feature;
	this.requestTag = tag;
	this.stopwatch = Stopwatch.StartNew();

	this.feature.UnregisterCancellableRequest(this.requestTag);

	req.Items[typeof(CancellableRequest).Name] = feature.RequestsMap[tag] = this;

	if (feature.OnRegisterCancellableRequest != null)
		feature.OnRegisterCancellableRequest(tag);
}
```

Nness:
If an operation is lengthy enough to require cancellation, shouldn't it be driven by [MQ](https://github.com/ServiceStack/ServiceStack/wiki/Rabbit-MQ)?

Or similar approach as [SSLabs](https://github.com/ssllabs/ssllabs-scan/blob/master/ssllabs-api-docs.md).

Basically, first call will initialize a task with id, they you use that id to track the progress, cancel if need or retrieve result if finish. Normal query shouldn't require cancellation and should fast.

drakelord:
MQ or not you need some way to cancel request. To support mq requests we just need to allow creating CancellableRequest from IMessage, which already has Tag and id.

You can use this feature to mimic SSLabs approach. In service start new Task and create CancellableRequest in it then just return response to client. Task will continue executing on server and you still can cancel it or even monitor its progress (with some modifications to this feature). If you need request's result you can store it in Cache and retrieve with another service call by request tag.

Feature modifications that allow client to pull request's progress status, but i think pushing progress updates from server via ServerEvents is better and can be done without this modifications.    
```c#
public interface ICancellableRequest : IDisposable
{
	CancellationToken Token { get; }
	CancellationTokenSource TokenSource { get; }
	TimeSpan Elapsed { get; }
	
	int Progress { get; set; }
	string ProgressStatusMessage { get; set;}
}

[Route("/current-requests/{Tag}/progress")]
[DataContract]
public class GetRequestProgress : IReturn<GetRequestProgressResponse>
{
	[DataMember(Order = 1)]
	public string Tag { get; set; }
}

[DataContract]
public class GetRequestProgressResponse
{
	[DataMember(Order = 1)]
	public string Tag { get; set; }

	[DataMember(Order = 2)]
	public TimeSpan Elapsed { get; set; }

	[DataMember(Order = 3)]
	public ResponseStatus ResponseStatus { get; set; }

	[DataMember(Order = 4)]
	public int Progress { get; set; }

	[DataMember(Order = 5)]
	public string ProgressStatusMessage { get; set; }

}

public object Any(GetRequestProgress request)
{
	if (request.Tag.IsNullOrEmpty())
		throw new ArgumentNullException("Tag");
	var cancallableReq = base.Request.GetCancellableRequest(request.Tag)
	if (cancallableReq == null)
		throw HttpError.NotFound("Request with Tag '{0}' does not exist".Fmt(request.Tag));

	return new GetRequestProgressResponse {
		Tag = request.Tag,
		Elapsed = cancallableReq.Elapsed,
		Progress = cancallableReq.Progress,
		ProgressStatusMessage = cancallableReq.ProgressStatusMessage
	};
}
```
Example usage.
```c#
public void Any(TestCancelRequest req)
{
	Task.StartNew(() => {
		using (var cancellableRequest = base.Request.CreateCancellableRequest())
		{
			while (true)
			{
				cancellableRequest.Token.ThrowIfCancellationRequested();
				Thread.Sleep(100);
				cancellableRequest.Progress += 100;
				cancellableRequest.ProgressStatusMessage = "Looping...";
			}
		}
	});
}
```

Nness:
One of downside of this approach is same user can accidentally start heavy task multiple times by simply hit that url. Heavy query should be safe if its execute at central dispatcher with monitor. So repeated request is ignored while different request are queued up on process.

So that server won't stressed up in any condition.

One of thing could be use task dataflow, either one dataflow per server application or per user. Of course there are other things have to resolve regard to monitor, cache etc.

I remember MSFT's media convert service in Azure had similar approach. They save progress status into their no-sql database. That supports load balance. When task start, they change DB, and update progress whenever need.

While SSLabs guard duplicate request well, since token is just your request, same request won't trigger another task.

drakelord:
This feature shouldn't guard you from multiple execution, you should do it yourself. For example I have some post request which scans directory on server and store file tree with hashes in db. 
```c#
public void Post(RehasDir req)
{
	var cache_tmp_key = string.Format("RehashTask:{0}", req.Dir);

	var val = Cache.Get<bool?>(cache_tmp_key);
	if (val.GetValueOrDefault())
	{
		throw new HttpError(HttpStatusCode.BadRequest, "400", string.Format("Rehash of {0} already running!", req.Dir));
	}
	Cache.Set<bool>(cache_tmp_key, true);
	try
	{
		// some code here
	}
	finally
	{
		Cache.Remove(cache_tmp_key);
	}
}
```
As you can see cache_tmp_key allows me to block only duplicated Dir request and allow all others.
I wanted to say, that decision which request is duplicate and which is not depends on your service implementation and must not be delegated to some global feature of framework.

Nness:
My concern is this shouldn't be framework's feature but implement individually. Current approach will block that request until its finished. (fix me if I'm wrong). All the example I have given, non of them block your call. All process are execute at background.

Cancellation can be done as long as you can find your process. Same as mythz's approach.

If you block request, depends on setting, it could time out before actually request is done. Make a request that will last long time doesn't feel right.

drakelord:
The last example indeed is blocking, but if you combine it with previous ones:
```c#
public void Post(RehasDir req)
{
    var cache_tmp_key = string.Format("RehashTask:{0}", req.Dir);

    var val = Cache.Get<bool?>(cache_tmp_key);
    if (val.GetValueOrDefault())
    {
        throw new HttpError(HttpStatusCode.BadRequest, "400", string.Format("Rehash of {0} already running!", req.Dir));
    }
    Cache.Set<bool>(cache_tmp_key, true);
	var cancelableRequest = base.Request.CreateCancellableRequest();
	Task.StartNew(() => {
        using (cancelableRequest)
        {
            try
			{
				// some code here
				cancellableRequest.Token.ThrowIfCancellationRequested();
			}
			finally
			{
				Cache.Remove(cache_tmp_key);
			}
        }
    });   
}
```
Long-running requests are bad, but sometime you need them. This feature allows you to track and cancel them in a simple and clean way, without overhead to other requests.

Nness:
This approach for immediately reply feels, I don't know. Cancel request feature cache IRequest in dictionary, while this approach use `Task.StartNew()` to put task into new thread. 

I believe this feature is good for some cases but not for other. Which is what it should be. Super machine is not good. Is it worth to have feature release as general? I don't know. For all long running tasks, in what percentage that it is good to use this feature? Hard to tell. Will  there be misuse of this feature? Definitely. 

I believe people should ask some questions to themselves before use this feature.
* How long will your request last?
* If request is eventually complete do you still want to cancel? In other word if there is background engine handle all tasks which guarantee it will complete without interrupt, do you still want to cancel it.
* Why do you want to cancel? because it last too long? or there is case where execution is wrong?
* Do you want client side to be that smart to have logic to cancel request? If it is how often do you update your client.
* Is task able to cancel? This goes with whether task is sync or async, if sync how often do you check cancellation token, if async, are there any async method that's not support token.
* How long will it take to cancel if you initialized it? 
* How heavy it is for long running task? Does it consume a lot of memory or CPU? Does it need max concurrency block for those tasks?
* Depends on how long it last and how often it run into multiple execution at same time, will there be impact to application's performance?

Because long running task should be properly evaluate before implement, I fear that people just pick up and use it without thinking. I might be worry too much. :P

AYss:
Hi everyone,

I guess it is my post no1, so: My name is Andy, but I go under AYss. I code for 24 years, in .NET for 10 years, in SS, RESTs and such for 1 year. I do not have any MVPs, MBAs, FTWs or WTFs, because I don't like 3-letter acronims.

Ok. Having formalities fullfilled I can say what I have in mind...

Nness: "Will there be misuse of this feature? Definitely."

You are 100% right. However, I believe that a good feature, cannot be 100% safe. It must have an area that have no boundries. Otherwise that would limit its usages to a concrete number of scenarios. Example: everyday you have an amazingly versile tool in your hand called a knife. It is so easy to misuse a knife...

Regarding the fuctionality itself, I have doubts if that's a way I would implement it. I would rather go towards using Id instead of Tags, maybe even implement Id into IRequest, as it is in IMessage, and reserve and use a thread on the server side for some light watchdog mechanism, that would ensure that something keeps an eye on server's interest.

But since I have no other concrete solution for the moment, I guess it's best we have. I assume it would not arise without some need, so there is a need for it, so unless anyone has a better idea, the above code is justified.

AY

PR #983, opened 2015-05-11
Fix wrong query/index on RavenDbUserAuthRepository

brunomlopes:
We just had a problem where a user named "user.name" was getting logged in as "user.name-special".

The problem is that the index used in raven to find the users matching a username or email analyses the username and email, and so it splits on some characters, including '-'.

Then the .Search method is used, which is used for full text search, where we want .Where, which is exact match.

This change fixes both issues.


mythz:
Cheers :+1:

PR #986, opened 2015-06-05
Applied fix for Session issues

balexandre:
A fix was applied as in several cases, the use of "using" will bubble up and close the database session and for such, resulting on an System.ObjectDisposedException Exception "Session is closed!" while performing normal tasks.


mythz:
Brilliant, thx!

PR #987, opened 2015-06-14
Log when errors are detected during application initialization

schmidt4brains:
1) Add the service name to initialization-complete message. This helps analysis when multiple logs from different processes get merged into a single stream.

2) Examine the StartupErrors collection to determine if anything went wrong during initialization.

Log.Info => when we know no errors occurred
Log.Error => when we know at least one error occurred. Also log the entire StartUpErrors collection as JSON, to help guide post-mortem analysis.


mythz:
Thx Doug :+1:

PR #988, opened 2015-06-24
AutoQuery with Aggregates

drakelord:
Hi @mythz !  This is more feature request, than actual pull request. It works, but i had some trouble with returning results. If you like this feature, I can rewrite this in more appropriate pull request for you to merge.

Main idea is to allow client to request additional aggregates and calculate them without extra call to DB.

Client can send query parameter Aggregates=SumAmount,MaxAmount ({Aggregate}{Field} convention).
Query is generated as usual, but instead regular execution SELECT COUNT(\*) is changed to 
SELECT COUNT(\*), SUM(Amount), MAX(Amount).

When it came to returning results, at first i used Dictionary\<string,object\> which worked with json format, but for obvious reasons, failed with ProtoBuf, Now its Dictionary\<string,string\>, maybe not very elegant, but should work almost everywhere.

Please review this code if you have time, thanks in advance.

### DIFF
--- a/src/ServiceStack.Server/AutoQueryFeature.cs
+++ b/src/ServiceStack.Server/AutoQueryFeature.cs
@@ -18,6 +18,10 @@ using ServiceStack.Web;
 using ServiceStack.Data;
 using ServiceStack.OrmLite;
 
+using ServiceStack.OrmLite.Dapper;
+using System.Collections.Concurrent;
+using System.Text;
+
 namespace ServiceStack
 {
     public delegate ISqlExpression QueryFilterDelegate(IRequest request, ISqlExpression sqlExpression, IQuery model);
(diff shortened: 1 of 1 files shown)


mythz:
Hi @drakelord 

Thanks, I like this feature although I wanted to expose it in a more familiar API, I also wanted to make it less coupled so that it can be more customizable and extensible so I've implemented it using AutoQuery ResponseFilters where users can pass in any number of commands that any Response Filter can access and populate the returned QueryResponse<T> with additional Metadata which I've added in this commit: https://github.com/ServiceStack/ServiceStack/commit/e43f1990cc2ec384b80829177c0bf00ccf779f56

I started with what I thought the ideal generic API for this functionality would look like, namely:

    ?include=Count(*)

Where the user can specify additional instructions of what they want added to the QueryResponse. I've decided to just go with SQL Syntax as it's more familiar and flexible and does a better job at describing how it works and what's happening. 

I've included some tests to show some examples of what's supported, e.g:

For the most basic usage you can just specify the aggregate function you want to include, e.g:

```csharp
var response = client.Get(new QueryRockstars { Include = "COUNT" });
```

When no arguments are supplied it defaults to `*` so the query ends up selecting `COUNT(*)`. I'm using the `Dictionary<string,string> Meta` to store the results which uses the column name as a key which is usually the same as the user supplied string but not always, e.g:

```csharp
response.Meta["COUNT(*)"]
```

It also supports SQL Column aliases, e.g:

```csharp
var response = client.Get(new QueryRockstars { Include = "Count(*) Total" });
response.Meta["Total"]

var response = client.Get(new QueryRockstars { Include = "Count(*) as Total" });
response.Meta["Total"]
```

As well as referencing fields:

```csharp
var response = client.Get(new QueryRockstars { Include = "MIN(Age)" });
response.Meta["MIN(Age)"]
```

Where it looks up to ensure it's a valid field, if it's not it's assuming you're passing in a string and quotes accordingly.

You can also use the `DISTINCT` modifier, e.g:

```csharp
var response = client.Get(new QueryRockstars { Include = "COUNT(DISTINCT LivingStatus) Total" });
response.Meta["Total"]
```

And just like yours supports multiple aggregates, e.g:

```csharp
var response = client.Get(new QueryRockstars { 
    Include = "Count(*) count, Min(Age) min, Max(Age) max, Sum(Id) sum" 
});
response.Meta["count"]
response.Meta["min"]
response.Meta["max"]
response.Meta["sum"]
```

The nice thing about this just being a response filter is that if anyone doesn't want this behavior they can remove it from the `AutoQueryFeature.ResponseFilters` list, replace it with their own impl or extend what's supported by adding their own Response Filter.

This is now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet) if you want to kick the tires and take it for a spin :)

drakelord:
Great, thx mythz, but with current implementation we end up with 3 query per call: total rows count, data select and aggregates. We can omit COUNT(\*) in Execute method and populate it from aggregates. It may seem, that one extra COUNT(\*) is not a big deal, but on large table and some complex filter it may double execution time. I don't see way to achieve this with ResponseFilters. Can you allow Execute method logic customization (maybe with some PreResponseFilters) and pass hole QueryResponse to ResponseFilter instead of meta?

mythz:
ok no worries although generally I don't consider an opt-in feature like this to be used enough to warrant forcing a coupled implementation but I've collapsed the query for Total into the Aggregate query in this commit: https://github.com/ServiceStack/ServiceStack/commit/cc1792987ced41720568052db467be479760ab1f

drakelord:
wow that was fast :) thx again mythz.

PR #989, opened 2015-06-28
Minor fixes in AutoQuery IncludeAggregates response filter.

drakelord:
1) clear limits and order by - we want to calculate aggregates for hole result not just current page and Order By will just fail in MS SQL without Group By on same fields
2) generate default alias if non supplied - MS SQL returns empty name for aggregates without alias

Current tests does not fail because of SqLite, they may require adjustments.


mythz:
Great, nice catch :+1:

PR #992, opened 2015-07-02
Adding broadcasting to RabbitMq

dittodhole:
Due to a SO question (http://stackoverflow.com/questions/23540107/broadcast-rabbitmq-messages-with-servicestack) I got inspired to adapt SS internally instead of doing all the stuff by hand as shown in my answer.

I am uncertain if I met all the prerequisite for a PR, but I've tried to stick to the rules...


mythz:
Awesome, looks like a cool feature, thx :+1: 

I'm currently preparing for the next release atm and wont have time to review this properly. I'll be able to go through and do a proper review sometime next week.

dittodhole:
No worries - glad to help and feedback always welcome!

mythz:
Hey don't think I'll take this implementation it's quite disruptive and adds a lot of overloads which isn't obvious to use so I want to look at a more declarative approach. 

Just want to be clear on what the use-case for this as it wasn't clear in the tests, is it just to change what the Response gets published to? i.e. to publish it to a fan out instead of a Response MQ?

dittodhole:
The change covers following points:
- Automatic creation of a fanout (d26c954)
- Publishing to an exchange with another routing key (ef5a119) than the queue name (to enable broadcasting without calling the base `Publish` directly), which internally uses the basic `Publish`
- Registering handlers to any queue by overcoming the constraints in the previous implementation on the type (f27c03a), which is the biggest change

Fair enough, casting is a big problem - but as no other MQ implementation features broadcasting that easily, I did not pull the methods to the interfaces (which one might consider).

The initial problem with `RegisterHandler` to be used with receiving brodcasts is, that handlers are bound to the type, and that there cannot be more than one handler per type per AppDomain (which might be an issue with brodcasting).
Another issue was that you cannot reuse a DTO for default publishing and broadcasting, as ... there's only one handler per type.
Another issue was that `RegisterHandler` took the routing key as the queue name.
So the only solution was to go for the *native* `BasicConsume`, which lacks retries, resilience, filters, and so forth.

I am not sure if you read the SO question, but basically the request is:
> Is there way to make method `myMessageService.CreateMessageQueueClient().Publish(myMessage);` broadcast messages to all receivers?

dittodhole:
To avoid casting I've added another interface layer.

mythz:
Sorry by declarative I meant that if a user would prefer to have the response broadcasted instead of being sent in the default `TResponse` **.inq** then ideally they would just need to do annotate their response with something like:

```csharp
[MqResponse(MqType.Broadcast)]
public class HelloResponse
{
    public string Result { get; set; }
}
```

Then each MQ Server could interpret that intent and implement it in the most appropriate way for their MQ server. e.g. Using a fanout for RabbitMQ. 

This way they don't need to learn any new API's or the underlying mechanics of how it works and the disruption to the existing code-base is minimal.

I'm going to look to see if how feasible this approach is when I have time.

dittodhole:
Yep, adding annotation instead of a `Broadcast` method seems legit.

The only problem is - and that is also the reason why the whole scenario is not *that* self-explanatory - the binding of the temporary queue to the fanout requires additional work as well.
Attaching all behaviour to an attribute might be a bit dangerous in case you want to reuse the DTO for broadcast and a dedicated queue - but I am also OK with stripping that possibility.

Nevertheless: I am a bit baffled ... actually it is not about responses, it is about sending a request/message :)

gabrielvdk:
+1, it would be great to have this kind of functionality with persisting queues so that even if your consumer is temporarily offline it will still receive the broadcasted messages.

theonlylawislove:
+1

mythz:
@theonlylawislove it's on hold until I get time to work on it, working on other higher priority stuff atm.

Formator:
++1
I would very like to have this feature. Is there any news when can we see it in SS?

mythz:
@Formator Wont be until after we ship [.NET Core support](https://servicestack.uservoice.com/forums/176786-feature-requests/filters/top) which will be our top priority from next release.

Formator:
@mythz Thank you for answer.

joelharkes:
I've been thinking about this for a while as well. I thought about the same declaration myth shows with the attribute, but is this really the way to go?

In a sense the question is this: who's responsibility should it be to determine if it's a queue message or a pub/sub message?

Like now in SS the message defines it's queue, but it can be overwritten the on the message publisher. Should this be the case for pub sub too?

MADCookie:
Bump.

Now that you have .NET Core released. Is this feature being considered to merge into the pipeline?

I would like this feature to support a pub / sub solution like @joelharkes

dittodhole:
Coming back after some time, I want to address my considerations regarding annotations:

I am no fan of "annotate the hell out of me", as annotations add a static (assembly) reference to DTO-projects - such references impose additional references to clients. I try to avoid annotations most of the time, to keep the interface (in dependency regards) as sleek as possible.

Therefore, I consider the registry-style a much more viable approach to such scenarios (ie fragment the logic (= how is the DTO mapped) away from the state (= DTO)), as a DTO should remain as dumb as possible = [Soc](https://en.wikipedia.org/wiki/Separation_of_concerns)

Additionally, I believe that simple publishing and broadcasting imply different semantics, which makes a low-level distinction a satisfactory approach.

Anyway, these are not my final thoughts - I just wanted to add some thoughts to the overall progress :)

Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress.

pAESSLAR:
Is anyone able to help?

https://github.com/ServiceStack/Issues/pull/113

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list 

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list 

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list 

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list 

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list 

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

mythz:
> Finally, I am actively thinking about closing this PR and implementing a contrib package, due to the sluggish overall (interest, discussion, and implementation) progress - and I do not exclude me, myself, and I from the blame list

Agreed. Feel free to continue discussion or post links to any external libraries on this thread.

dittodhole:
@mythz Thanks for your follow-up, no offense intended or taken :neckbeard:

PR #993, opened 2015-07-08
avoid NRE when setting null value to cache

migajek:
Setting null-value to ICacheClient caused NRE when using Memcached backend


mythz:
Cheers!

PR #994, opened 2015-07-08
Swagger ApiMemberAttribute DataType 

stout01:
The problem that I was having was that the ApiMemberAttribute down the object chain inside Employee would not be picked up by swagger, such as in the example situation below.

The changes committed here add the DataType from ApiMemberAttribute if it exists.

```C#
public class SaveEmployee : IReturn<SaveEmployeeResponse>
{
       [ApiMember(IsRequired = true, ParameterType = "path", AllowMultiple = false, DataType = "int")]
        public string id { get; set; }

        [ApiMember(ParameterType = "body", IsRequired = true, AllowMultiple = false, DataType = "Employee")]
        public Employee employee { get; set; }
}

public class Employee
{
        [ApiMember(IsRequired = true, AllowMultiple = false, DataType = "int")]
        public string id { get; set; }
}
```

image

### DIFF
--- a/src/ServiceStack.Api.Swagger/SwaggerApiService.cs
+++ b/src/ServiceStack.Api.Swagger/SwaggerApiService.cs
@@ -378,10 +378,16 @@ namespace ServiceStack.Api.Swagger
                 else
                 {
                     ParseModel(models, propertyType, route, verb);
+
+                    var propAttr = prop.AllAttributes<ApiMemberAttribute>().ToList();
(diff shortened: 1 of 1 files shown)


mythz:
Thanks for PR, Could you also [approve the Contributor License Agreement](https://docs.google.com/forms/d/16Op0fmKaqYtxGL4sg7w_g-cXXyCoWjzppgkuqzOeKyk/viewform) so we can merge this, thx.

stout01:
Sorry about that.  I have now filled out the Contributor License Agreement.

mythz:
Cheers :+1:

PR #995, opened 2015-07-15
generate suffix if no alias supplied  

drakelord:
missed during refactoring c4f8af70e6aa51313d89807940d7d13abaa87710

### DIFF
--- a/src/ServiceStack.Server/AutoQueryFeature.cs
+++ b/src/ServiceStack.Server/AutoQueryFeature.cs
@@ -320,6 +320,10 @@ namespace ServiceStack
 
                     cmd.Suffix = " " + alias.SafeVarName();
                 }
+                else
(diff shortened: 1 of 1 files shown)


mythz:
Do you have an example of why we need an alias?

drakelord:
Try executing multiple aggregates without aliases against MS-SQL and Auto Query will return Meta with one empty key. This happens because MS-SQL returns empty column name for aggregate without alias and same key gets overwritten in meta. Current tests pass because they are executed against SQLite which behaves as AutoQuery implementation expects (e.g. returns COUNT(*) as column name).
Moreover, since auto-generated COUNT(*) cannot be picked from meta and its executed again to fill Total.

mythz:
sweet thx, all good reasons :)

PR #996, opened 2015-07-19
FluentValidation: Pass CustomState all the way down to the ResponseError

migajek:
This allows the clients to access the custom state along with message.
For me this is must-have; I'm using ErrorCode only to display custom (localized) message in the client.
This way I can pass additional argument, please consider the custom "length" rule(s):

```C#
public static class MyValidatorExtensions
    {
        public static IRuleBuilderOptions<T, string> LengthWithState<T>(this IRuleBuilder<T, string> ruleBuilder,
            int exactLength)
        {
            return ruleBuilder.Length(exactLength).WithState(x => new {Exact = exactLength});
        }

        public static IRuleBuilderOptions<T, string> LengthWithState<T>(this IRuleBuilder<T, string> ruleBuilder,
            int min, int max)
        {
            return ruleBuilder.Length(min, max).WithState(x => new {Min = min, Max = max});
        }
    }
```

that way I can build custom error message containing info on length constrainment(s)


migajek:
Accidentally  included changed ``ServiceStack.Interfaces.csproj`` (disabled signing), sorry!!

mythz:
Hi Michał, this PR cannot be cleanly merged, please make sure you pull changes from master before sending PR's. 

Also don't add `object` properties to any built-in core DTO's it breaks metadata support and [causes several issues for deserialization requiring serialization-specific hints](http://stackoverflow.com/a/10759250/85785) which couples the serialization output to specific concrete Serializers. You can use the `IMeta` interface and Dictionary instead for extending existing DTO's with custom info.

PR #997, opened 2015-07-19
FluentValidation: Pass CustomState all the way down to the ResponseError (fixed)

migajek:
I've fixed the code according to your guidelines (https://github.com/ServiceStack/ServiceStack/pull/996)


mythz:
Can you also rename it to `Meta` so it matches the field it's serialized as. and change it to a string Dictionary so it maps cleanly to what gets serialized. (i.e. so it's not unknowingly serialized by the framework).

mythz:
Actually don't worry about it, the dirty commits needed to be squashed with a rebase. I'll try add the feature manually by pulling it from the commits.

mythz:
I've added the extra Meta fields in this commit: https://github.com/ServiceStack/ServiceStack/commit/87d6fea092094414485d3d1e1405ee52ab858459

I left `ValidationFailure.CustomState` as it is since that's what FluentValidation uses but it needs to be a `Dictionary<string,string>` for it to be serialized to the error response dtos.

migajek:
Thanks Demis and sorry for the trouble, hopefully I'll do better with the code quality next time :)

mythz:
Yep no worries :)

PR #999, opened 2015-08-02
Bug fix: Inconsistent usage of DateTime.Now and .UtcNow in RabbitMqQueueClient

gabrielvdk:
Fixed bug in RabbitMqQueueClient that caused the time out based Get<T> to do an invalid calculation of the time span. It would always do a comparison between a DateTime.UtcNow based start time and a DateTIme.Now (not UTC) current time. This meant that on all machines not running on UTC time the initial difference would always be the difference between Utc-time and the time of time zone configured on the host machine.

### DIFF
--- a/src/.vs/config/applicationhost.config
+++ b/src/.vs/config/applicationhost.config
@@ -163,7 +163,7 @@
             </site>
             <site name="ServiceStack.WebHost.IntegrationTests" id="2">
                 <application path="/" applicationPool="Clr4IntegratedAppPool">
-                    <virtualDirectory path="/" physicalPath="C:\src\ServiceStack\tests\ServiceStack.WebHost.IntegrationTests" />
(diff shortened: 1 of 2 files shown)


mythz:
Hi thx for the fix :+1: 

I've committed in manually in https://github.com/ServiceStack/ServiceStack/commit/69f6b877aadabb9698d288476cb912ea27b22b08 as this PR included local path changes to `/.vs/config/applicationhost.config`, which have since been ignored: https://github.com/ServiceStack/ServiceStack/commit/b6751ab54020d3c0a0c86b4e0f2d7050c4060fa1 

Please also [approve the Contributor License Agreement](https://github.com/ServiceStack/ServiceStack/commit/69f6b877aadabb9698d288476cb912ea27b22b08) so we're able to merge future PR's.

PR #1000, opened 2015-08-07
Don't create SessionId's with Url unfriendly chars

kinstephen:
Merging v4 fix to the v3 branch for unfriendly url chars in the
sessionid.


mythz:
Cheers :+1:

PR #1003, opened 2015-09-07
Postman responses can't be null

balexandre:
From a support email, Postman expects the "responses" to be an empty array if empty, but never "null" or the import collection in the app will throw an error.
This prevents the responses property to be null and if empty, it will output an empty array instead.

### DIFF
--- a/src/ServiceStack/PostmanFeature.cs
+++ b/src/ServiceStack/PostmanFeature.cs
@@ -72,6 +72,11 @@ namespace ServiceStack
 
     public class PostmanRequest
     {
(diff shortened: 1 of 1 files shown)


mythz:
Great thx Bruno, Could you also approve the [Contributor License Agreement](https://docs.google.com/forms/d/16Op0fmKaqYtxGL4sg7w_g-cXXyCoWjzppgkuqzOeKyk/viewform) so we can merge this, Cheers.

mythz:
Cheers :+1:

PR #1004, opened 2015-09-21
Added Patch method to JsonServiceClient.js and JsonServiceClient.closure.js

DeonHeyns:
I needed to add the Patch verb for a project I am working on. I couldn't find any tests for the JsonServiceClient.js to update with this change. Also not sure if this is the right place to contribute this change as the JsonServiceClient.js and JsonServiceClient.closure.js was in the lib folder?


mythz:
Great thx, nice to see the `JsonServiceClient` being used :) most js devs just use call $.ajax directly.

DeonHeyns:
Thanks for accepting @mythz!

I stumbled acrossed it on the [Javascript client](https://github.com/ServiceStack/ServiceStack/wiki/Javascript-client) page. Would you mind if I added an example of using it? Think it would help others and might encourage usage.

mythz:
Sure go ahead,  would love to see more examples :)

PR #1005, opened 2015-10-01
proposing an annotation to indicate that on insert, the tagged proper…

yuinlin:
…ty is always set to UTC from database server
the attribute will indicate to ormlite to generate insert statement with embedded database utc function for the property, instead of using the property value (if any) .  for example, for an entity like
```
public class Test
        {
            [PrimaryKey]
            public int Id
            {
                get;
                set;
            }            
            [InsertDatabaseUtcAttribute]
            public DateTime CreatedTimeUtc
            {
                get;
                set;
            }
        }
```
the insert statement would be
```
oracle- insert into Test (Id, CreatedTimeUtc) values (1, sys_extract_utc(systimestamp))
sqlserver - insert into Test (Id, CreatedTimeUtc) values (1, SYSUTCDATETIME())
mysql - insert into Test (Id, CreatedTimeUtc) values (1, UTC_TIMESTAMP())
```
etc

please let me know what you think, thanks

### DIFF
--- a/src/ServiceStack.Interfaces/DataAnnotations/InsertDatabaseUtcAttribute.cs
+++ b/src/ServiceStack.Interfaces/DataAnnotations/InsertDatabaseUtcAttribute.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace ServiceStack.DataAnnotations
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class InsertDatabaseUtcAttribute : AttributeBase
+    {
+    }
+}
--- a/src/ServiceStack.Interfaces/ServiceStack.Interfaces.csproj
+++ b/src/ServiceStack.Interfaces/ServiceStack.Interfaces.csproj
@@ -91,6 +91,7 @@
     <Compile Include="DataAnnotations\IgnoreAttribute.cs" />
     <Compile Include="DataAnnotations\PrimaryKeyAttribute.cs" />
     <Compile Include="DataAnnotations\RequiredAttribute.cs" />
+    <Compile Include="DataAnnotations\InsertDatabaseUtcAttribute.cs" />
     <Compile Include="DataAnnotations\RowVersionAttribute.cs" />
     <Compile Include="DataAnnotations\SchemaAttribute.cs" />
     <Compile Include="DataAnnotations\SequenceAttribute.cs" />


jklemmack:
How is this different than specifying a default formula on the table definition, and submitting a null value?

yuinlin:
it removes the need to "hide" properties in order to ensure a null value

mythz:
Don't think this Attribute should be in OrmLite, it's very specific where I don't want for this to be how we'd start to add data type specific features like this. Custom attributes with specific behavior are not discoverable and has limited applicability so I'd expect this would hardly ever be used so don't think it should be in OrmLite.

It's cleaner for this to a default value or a trigger. OrmLite already has support for default values and custom fields where you should be able to specify it in the CREATE TABLE statement with:

```csharp
public class Test
{
    [Default("SYSUTCDATETIME()")]
    public DateTime CreatedTimeUtc { get; set; }

    [CustomField("DATETIME2(7) DEFAULT (SYSUTCDATETIME())")]
    public DateTime CreatedTimeUtc2 { get; set; }
}
```

I've added support for variable placeholders in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/0e6d55de083d1e72b29ad2fd5b94a23d27507a2f

Where you can have a multi-database solution with the `{SYSTEM_UTC}` placeholder that each provider has a different value for, so you can now specify a default System UTC value with:

```csharp
public class Test
{
    [Default(OrmLiteVariables.SystemUtc)]
    public DateTime CreatedTimeUtc { get; set; }
}
```

I've also modified `db.Insert()` to no generate parameters for fields annotated with `[Default]` that also have a default value, so the default value should now be populated with `DateTime` value types.

yuinlin:
@mythz thanks for ServiceStack/ServiceStack.OrmLite@0e6d55d, very helpful.  i think there's an edge case in PrepareParameterizedInsertStatement- if all columns for a table are "skipped", the resulting insert statement would have an invalid empty column list.  i stumbled on this when i added auto increment to the id property
```
    public class DefaultValues
    {
        [AutoIncrement]
        public int Id { get; set; }

        [Default(1)]
        public int DefaultInt { get; set; }

        [Default(1)]
        public int? NDefaultInt { get; set; }

        [Default(1.1)]
        public double DefaultDouble { get; set; }

        [Default(1.1)]
        public double? NDefaultDouble { get; set; }

        [Default("'String'")]
        public string DefaultString { get; set; }

        [Default(OrmLiteVariables.SystemUtc)]
        public DateTime CreatedDateUtc { get; set; }

        [Default(OrmLiteVariables.SystemUtc)]
        public DateTime? NCreatedDateUtc { get; set; }
    }
```
don't know if this is a likely use case.

mythz:
Yeah that would be an edge case that we wouldn't be able to support, there needs to be at least 1 field in the INSERT statement.

PR #1006, opened 2015-10-10
Propose serializing directly into compression stream when making compressed result

BlakeRaymond-AI:
Avoid storing the uncompressed response as a string in memory; serialize into a GZip/Deflate stream directly. We have implemented this in a custom service runner for very large responses (hundreds of MB uncompressed) that compress well for a much healthier memory footprint. 

Please let me know what you think. Thanks.

### DIFF
--- a/src/ServiceStack.Client/Support/NetDeflateProvider.cs
+++ b/src/ServiceStack.Client/Support/NetDeflateProvider.cs
@@ -34,6 +34,10 @@ namespace ServiceStack.Support
             }
         }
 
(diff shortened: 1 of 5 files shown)


mythz:
Yep change looks good, thx :+1: 

but I've renamed the interface methods to DeflateStream / GZipStream which is a more natural and predictive naming relative to the current APIs:
https://github.com/ServiceStack/ServiceStack/commit/538555a91eaa4259d76ed72ead5d4ca8e8bff0fa

BlakeRaymond-AI:
Thanks!

PR #1007, opened 2015-10-22
Fixed bug regarding unauthorised errors returning 302 instead of 401 when HtmlRedirect is null

bumblebeeman:
As discussed [here](https://github.com/ServiceStack/Issues/issues/363)

### DIFF
--- a/src/ServiceStack.Mvc/ServiceStackController.cs
+++ b/src/ServiceStack.Mvc/ServiceStackController.cs
@@ -41,7 +41,7 @@ namespace ServiceStack.Mvc
         /// </summary>
         public virtual string UnauthorizedRedirectUrl
         {
-            get { return HostContext.GetPlugin<AuthFeature>().GetHtmlRedirect() + "?redirect={0}#f=Unauthorized"; }
+            get { return HostContext.GetPlugin<AuthFeature>().GetHtmlRedirect(); }
         }
 
         /// <summary>
@@ -53,7 +53,17 @@ namespace ServiceStack.Mvc
             get
             {
                 var returnUrl = HttpContext.Request.GetPathAndQuery();
-                return new RedirectResult(UnauthorizedRedirectUrl.Fmt(returnUrl.UrlEncode()));
+
+                var encoded = returnUrl.UrlEncode();
+                var unauthorizedRedirectUrl = UnauthorizedRedirectUrl;
+
+                if (string.IsNullOrEmpty(unauthorizedRedirectUrl))
+                {
+                    throw new HttpException(401, "Unauthorized");
+                }
+
+                var formatted = unauthorizedRedirectUrl.Fmt(encoded);
+                return new RedirectResult(formatted);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/ServiceStack.Mvc/ServiceStackController.cs
```diff
@@ -229,10 +250,6 @@ protected virtual TUserSession SessionAs<TUserSession>()
         {
             return ServiceStackProvider.SessionAs<TUserSession>();
         }
-        protected virtual void SaveSession(IAuthSession session, TimeSpan? expiresIn = null)
-        {
-            ServiceStackProvider.Request.SaveSession(session, expiresIn);
-        }
```
bumblebeeman: Was working of code from a couple of days ago, I'm guessing you added this after

src/ServiceStack.Mvc/ServiceStackController.cs
```diff
@@ -53,7 +53,17 @@ public virtual ActionResult AuthenticationErrorResult
...
+
+                var encoded = returnUrl.UrlEncode();
+                var unauthorizedRedirectUrl = UnauthorizedRedirectUrl;
+
+                if (string.IsNullOrEmpty(unauthorizedRedirectUrl))
+                {
+                    throw new HttpException(401, "Unauthorized");
+                }
+
+                var formatted = unauthorizedRedirectUrl.Fmt(encoded);
```
Nness: You removed `?redirect={0}#f=Unauthorized`, where you using string.format encoded after.. Even though no error will be throwing, but what's the point? And by this change you removed default behavior. 

Things can be improve is under AuthFeature class's GetHtmlRedirect.
```
if (feature != null)
                return feature.HtmlRedirect;
```
to 
```
if (feature != null && feature.HtmlRedirect != null)
                return feature.HtmlRedirect;
```
Also HtmlRedirect did setup automatically, unless you change that value after AuthFeature been created.

bumblebeeman: Ok, I did raise this as an issue first, as I wasn't sure.

The main point is, in the documentation it says that if you set HtmlRedirect to null, you will get a 401 response when unauthorised, instead you get a 302, which I don't want.

I only want a 302 when I set HtmlRedirect to an actual value, with the temporary redirection being to that value.

Nness:
Well I'm not author, just curious about this pull request.

I misunderstand part of your fix I think. My proposal is wrong.

One way of fix without changing behavior might be following. 

```
        public virtual ActionResult AuthenticationErrorResult
        {
            get
            {
                var returnUrl = HttpContext.Request.GetPathAndQuery();
                var unauthorizedRedirectUrl = UnauthorizedRedirectUrl;
                if (unauthorizedRedirectUrl.IsNullOrEmpty() || unauthorizedRedirectUrl.StartsWith("?redirect")) {
                    return new HttpUnauthorizedResult("Unauthorized");
                }
                return new RedirectResult(unauthorizedRedirectUrl.Fmt(returnUrl.UrlEncode()));
            }
        }
```

What do you think? This way it won't change original behavior. of `UnauthorizedRedirectUrl`.

bumblebeeman:
Ok, I guess I will wait and see if this gets accepted as a pull request, and if not submit another one, it's a fairly minor change so might be easier for the devs (mythz?) to pull and then patch.

One thought, I prefer throwing HttpExceptions, as there is no *HttpForbiddenResult*, so this keeps it consistent with the code below.

mythz:
Hey everyone I've just added a commit to clean up the PR with both your suggestions, please let me know if there are other changes you would like to make.

PR #1008, opened 2015-10-22
Authenticate override for Instagram no longer needed

brunomlopes:
A recent change in Authenticate handled the return value of OnAuthenticated as the return value, but that behaviour was missing on this provider and broke a particular flow on our app.

A quick comparisson between the base provider and this one seemed to show that the only relevant difference could be handled by AuthClientFilter.

I'm not 100% sure on this, but thought it's easier to show the change this way.


mythz:
Cool, I like it, deleting code is a win! thx :+1:

PR #1009, opened 2015-11-04
Proposing an extension to Default attribute

BruceCowan-AI:
Rationale: When using the Default attribute, especially with arguments such
as OrmLiteVariables.SystemUtc, it is frequently true that you want the default
value set on an update. Currently the only way to do that in OrmLite is to
use custom SQL.

You may remember https://github.com/ServiceStack/ServiceStack/pull/1005 which
you rejected in favor of the Default attribute. So we are using the Default
attribute and finding that it needs update functionality. Hence this PR and if
you approve, the actual work to implement it in OrmLite.

The "UseOnUpdate=true" extension would be implemented by each dialect provider
by either generating SQL to set the column value on an update operation or by
creating a trigger which does it. For an UpdateOnly operation, any UseOnUpdate
columns would only be updated if the onlyFields expression includes them. For
an UpdateNonDefaults operation UseOnUpdate columns would not be updated.

### DIFF
--- a/src/ServiceStack.Interfaces/DataAnnotations/DefaultAttribute.cs
+++ b/src/ServiceStack.Interfaces/DataAnnotations/DefaultAttribute.cs
@@ -11,8 +11,11 @@ namespace ServiceStack.DataAnnotations
         public Type DefaultType { get; set; }
         public string DefaultValue { get; set; }
 
+        public bool UseOnUpdate { get; set; }
+
         public DefaultAttribute(int intValue)
         {
+            this.UseOnUpdate = false;
             this.IntValue = intValue;
             this.DefaultType = typeof(int);
             this.DefaultValue = this.IntValue.ToString();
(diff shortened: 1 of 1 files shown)


Siliconrob:
Question if you want the default value provided by the specific backing DataStore why not fix the DataStore behavior i.e. in Sql Server you could use a stored procedure or a default value condition in the DataStore behavior?

mythz:
Hi Bruce, fine with the addition but IMO `OnUpdate` is a better name since `Use` is implied and when unambiguous, less is more in naming. I've added the change in this commit: https://github.com/ServiceStack/ServiceStack/commit/08a5dcd4485c44e78013c28c21e1d4b2af6ae484

The boolean field doesn't need initializing since it defaults to false.

I've also updated the latest ServiceStack.Interfaces.dll to SS and SS.OrmLite repos.

BruceCowan-AI:
@mythz Thanks Demis.

@Siliconrob For sure, the implementation can use any of stored procedures, triggers, default value conditions, or put it into the SQL generated for the update. This change is just about giving the implementation the information necessary for it to generate the correct update behavior. Since we haven't started implementing yet, I don't know which we'll use in which provider.

PR #1011, opened 2015-12-03
Suggestion for generated help message

Layoric:
Added a `Tip` to generated native types to help devs with convention of overriding DTO options by removing comment prefix.

Idea from [SSVS raised issue](https://github.com/ServiceStack/ServiceStackVS/issues/21).

Thoughts?


mythz:
Don't think it's required, it is intuitive as this is how any config file works and it's all over the docs with examples. It should take more than one person who didn't read the docs to make changes like this that's visible for everyone. But I think we can carry this for a while although I'd look at remove it later once everyone gets more familiar with it as IMO it's unnecessary.

Layoric:
Yeah I agree, only one other example I think it has tripped someone up.

https://github.com/ServiceStack/Issues/issues/274#issuecomment-93892640

Also first example of non data driven content in the result which looks out of place.

PR #1012, opened 2015-12-10
Change ResponseStatus type to ResponseStatus?

Layoric:
For null assignment. Only affects classes without ResponseStatus already declared in response and `AddResponseStatus: True` in options.


mythz:
Cheers :+1:

PR #1014, opened 2015-12-19
Use ConcurrentDictionary to store IEventSubscription instead of Array[]

Nness:
Our SSE server had a bit of trouble lately. The week before, our server become exhausted around 2am NZ time every day. Here is detail of what happen.

Our instance: AWS 8 cores instance with enough ram.
We have average 25k ~ 30k subscriptions during off peak time.

*** Reason drive me to made those change: *** We have been keep creating perfview for our SSE server, what I have observed was high CPU usage during server restart. Which has higher number of subscription came back in. It is similar event as the peak time. The error client side get is 504 service not available, where server is s till functional.

From perfview, during server restart, highest server usage was on `Array.Copy` and ConcurrentDictionary's `TryGetValue`. Where they are all under Registry method.

Before 14th, server keep became exhausted around 2am NZ time during our server's peak time.

14th of Dec:
Clone our SSE server and off load around 5k~8k subscriptions to the new box.
Old SSE server upgrade with the change I made to this pull request.

15th of Dec:
Original SSE server didn't go down during peak hour. CPU and Memory stay low.
New SSE server with only 1/4 of subscriptions become exhausted require restart IIS.

16th of Dec:
Same as 15th

17th of Dec:
We decide to use my new build on new SSE server. Which result as what we expected. CPU remain around 5~10 percent throughout 17th and 18th.

 Here is quote from my manager when he was checking difference during 16th of Dec's event:
```
SSE2 server went down last night. As comparison at the same time of day when SSE2 was consuming 63% of CPU and 1.29 MB/s network, SSE1 only consume 19% of CPU and 4.12 MB/s.
SSE2 been hit with 1/3 of the network I/O but is using 3x the CPU.
Event at 10MB/s on SSE1 it was only 27.6 CPU
```

### Our new relic graph for SSE1 server before change. 
pasted image at 2015_12_18 10_16 am
pasted image at 2015_12_18 10_17 am

### New relic graph for SSE1 server after change
pasted image at 2015_12_18 10_14 am
pasted image at 2015_12_18 10_15 am

### DIFF
--- a/src/ServiceStack/ServerEventsFeature.cs
+++ b/src/ServiceStack/ServerEventsFeature.cs
@@ -764,6 +764,7 @@ namespace ServiceStack
                     }
                     RegisterSubscription(subscription, subscription.SubscriptionId, Subcriptions);
                     RegisterSubscription(subscription, subscription.UserId, UserIdSubcriptions);
+                    RegisterSubscription(subscription, subscription.UserName, UserNameSubcriptions);
                     RegisterSubscription(subscription, subscription.SessionId, SessionSubcriptions);
 
                     if (OnSubscribe != null)
--- a/src/ServiceStack/ServerEventsFeature.cs
+++ b/src/ServiceStack/ServerEventsFeature.cs
@@ -1010,12 +1010,19 @@ namespace ServiceStack
 
         internal static IEnumerable<TElement> ValuesWithoutLock<TKey, TElement>(this ConcurrentDictionary<TKey, TElement> source)
         {
-            return source.Skip(0).Select(item => item.Value);
+            foreach (var item in source)
+            {
+                if (item.Value != null)
+                    yield return item.Value;
+            }
         }
 
         internal static IEnumerable<TKey> KeysWithoutLock<TKey, TElement>(this ConcurrentDictionary<TKey, TElement> source)
         {
-            return source.Skip(0).Select(item => item.Key);
+            foreach (var item in source)
+            {
+                yield return item.Key;
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


mythz:
Thanks for the detailed stats and PR. Am I reading this right that you've only been running this new impl for a couple of days? in which case can you let me know how it goes after a couple more days of usage?

Also it looks like the memory usage is up with the new impl, do you think that's due to the change of using ConcurrentDictionary over arrays? or did it have more connections?

Nness:
Our main SSE server been running this build for a week now I believe. I will check again tomorrow when I come back to the office. 

For memory consumption, it is hard to say, it could be both. ConcurrentDictionary definitely will consume more memory than array, but how much more it is question. I will take memory dump tomorrow and take look. And I can put somewhere if you are interested as well.

Nness:
One thing to be in mind is, we have to restart IIS each day before new implementation, and we use RavenDB. RavenDB client has cache data of 1000, as the application running longer than 1 day, cache data could be more than the one within 1 day. That's just one of possibility, It will be interesting to see what actually holds.

Nness:
Just did memory dump and here is result. At the time I made memory dump, subscriptions count was around 10k+.

2015-12-21_11h23_46
2015-12-21_11h23_58
2015-12-21_11h24_08

Nness:
I have also changed my existing code to not using Linq inside memory event, after deploy memory hit 3GB while subscriptions count was around 11k.

After few hours memory drop to 2GB with similar count as 11k.

Then half hour later with 13k subscriptions. Memory rise to 2.8GB.

I believe its IIS that's managing memory, hardly the case for ConcurrentDictionary to taken 1~2GB of ram.

mythz:
ok np thx for the info, hopefully it stays within normal range and doesn't cause perf regressions. I'll run with it since it looks like it solves your load issue.

thx for the PR!

Nness:
Cool.

As for that memory dump, when I took it, it was 4GB inside memory and 5GB as dump file.

PR #1015, opened 2016-01-04
Return all request binding errors with serialization exception

georgehemmings:
As details of the request binding error are now returned, I've extended this to return all binding errors, rather than just the first one. So if a field fails to bind the process continues to try to bind the other fields, rather than throwing. The end result is that all binding errors should be returned to the client in one go, which is particularly desirable if the errors are indicated to an interactive user for correction.

Related: https://forums.servicestack.net/t/different-binding-behavior-based-on-content-type/1681


mythz:
Awesome, looks good thx George :+1:

PR #1016, opened 2016-01-05
Update OrderedDictionary.cs

wulfgar61:
We are enforcing SonarQube static code analysis against our code.  It has recommended changes to the two lines I altered.  

Here is the rule we are trying to satisfy with the change.
When constraints have not been applied to restrict a generic type parameter to be a reference type, then a value type, such as a struct, could also be passed. In such cases, comparing the type parameter to null would always be false, because a struct can be empty, but never null. If a value type is truly what's expected, then the comparison should use default(). If it's not, then constraints should be added so that no value type can be passed.

### DIFF
--- a/src/ServiceStack.Api.Swagger/Support/OrderedDictionary.cs
+++ b/src/ServiceStack.Api.Swagger/Support/OrderedDictionary.cs
@@ -403,7 +403,7 @@ namespace ServiceStack.Api.Swagger.Support
         /// <remarks>This method performs a linear search; therefore it has a cost of O(n) at worst.</remarks>
         public int IndexOfKey(TKey key)
         {
-            if (null == key)
+            if (EqualityComparer<TKey>.Default.Equals(key, default(TKey)))
                 throw new ArgumentNullException("key");
 
             for (int index = 0; index < List.Count; index++)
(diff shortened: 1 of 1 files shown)


mythz:
Hi can you [approve the Contributor License Agreement](https://docs.google.com/forms/d/16Op0fmKaqYtxGL4sg7w_g-cXXyCoWjzppgkuqzOeKyk/viewform) so we can merge this, thx.

wulfgar61:
Done, thanks

mythz:
Great, thx :+1:

Nness:
Just curious. If @wulfgar61's change is correct then does MSFT have to change their code too?
[Dictioanry Insert Exception on Null](http://referencesource.microsoft.com/#mscorlib/system/collections/generic/dictionary.cs,321)

The thing struggling me is the comparer is not even using the comparer that manually passed into OrderedDictionary. So if I create a custom IEqualityComparer and return false when compare to default of that value. This will  still throw ArgumentNullException?

And not saying throwing ArgumentNullException is so weird on struct and not commonly expected.

mythz:
@Nness thx for catching and reporting this issue at: https://github.com/ServiceStack/Issues/issues/388 which shows that this change introduced bugs so I've reverted this change at: https://github.com/ServiceStack/ServiceStack/commit/752178da29e293907d9f2e50cb0710345d807e54

It should be ok to have a key with default value type whilst guarding against null reference type as the existing code does.

PR #1017, opened 2016-01-22
Fix SSE client restart

Layoric:
Currently when updating `Channels` of an SSE client, restarting didn't update the `EventStreamUri`. Test and fix in PR.

I could be missing the expected flow, but fix was due to use case of:

```` csharp
var channels = new string[] { "home" };
var client = ServerEventsClient(baseUrl,channels);
client.Start();
client.Channels = new string[] { "new-channel" };
client.Restart();
// Excepted client to be listening on updated channels only.
````


Layoric:
Oops, my mistake, should have been private. PR just trying to sort out `EventStreamUri` state to be accurate after restart. Updated PR.

PR #1018, opened 2016-02-05
Change cast of ServiceClient in SSE client

Layoric:
To support other IServiceClients that implement IHasCookieContainer, eg
JsonHttpClient.


mythz:
Yep cool, thx.

PR #1022, opened 2016-03-29
Allow custom mapping of C# Exceptions to Error Codes

georgehemmings:


src/ServiceStack/HttpRequestExtensions.cs
```diff
@@ -317,6 +317,16 @@ public static int ToStatusCode(this Exception ex)
 
         public static string ToErrorCode(this Exception ex)
         {
+            if (HostContext.Config != null)
+            {
+                var exType = ex.GetType();
+                foreach (var entry in HostContext.Config.MapExceptionToErrorCode)
+                {
+                    if (entry.Key == exType)
```
georgehemmings: As error codes tend to be more specific than HTTP Status Codes I think subclasses should be explicitly customized.

mythz:
Going with a custom `IHasErrorCode` interface instead of a static dictionary since it's less important than mapping to HTTP Status Codes which doesn't justify polluting the configuration with a custom dictionary that will ever rarely be used also static strings don't go far enough for those who do want to customize error codes who may want to use the same exception type but unique error code numbers, so I'm going with a custom `IHasErrorCode` interface which allows greater flexibility, "no touch" configuration and requires less code, added in this commit: https://github.com/ServiceStack/ServiceStack/commit/e8cdb55258164de7ee45cee2e7095e08d20e1d32

georgehemmings:
This is great for custom exceptions, but for BCL exceptions there isn't a simple way to override the ErrorCode. However, I'm starting to think I'll be better off using custom exceptions for those I want to bubble back to the client anyway. The flexibility of using `IHasErrorCode` to provide unique error codes will be useful, thanks for adding this feature.

mythz:
For custom BCL exceptions you can also wrap the Exception in a `HttpError` but yeah if you want a Custom ErrorCode it's better to go with a Custom Exception.

PR #1023, opened 2016-04-05
Instagram profile image is empty

kirankc:
When you authenticate using Instagram, the profile image is not showing up. Fixing code so that the profileUrl field gets added to items collection thereby causing the image to show up.


mythz:
Great, thx :+1:

PR #1025, opened 2016-04-25
Move Slack logging to SS logging sln

Layoric:
Add documentation and comments.
Build still doesn't work locally due to signing key required.

Tests required core ServiceStack library as I was bringing up an self host app to collect the requests that would otherwise be sent to Slack and checking if the payloads were correct.

See [here](https://github.com/Layoric/ServiceStack.Logging.Slack) for tests working.

Not sure how to best incorporate these kinds of tests with this solution seeing it uses external project references etc and ServiceStack.Interfaces PclDebug build which I can't build without singing key. @mythz could you have a look to see if you can build?


mythz:
Cool I'll have a look, note you shouldn't have to build ServiceStack.Interfaces project and can instead access the pre-built `lib/ServiceStack.Interfaces.dll`.

PR #1031, opened 2016-05-12
Allow ServerEventsClient to change baseUri.

Nness:
Add ability to change baseUri for `ServerEventsClient`. Restart connection is required after change baseUri. Which can be performed manually.

### DIFF
--- a/src/ServiceStack.Client/ServerEventsClient.cs
+++ b/src/ServiceStack.Client/ServerEventsClient.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


mythz:
What's an example use-case of why this would be necessary?

In future please also include tests that tests any new features.

Nness:
Our architecture is a bit different to normal situation, where we have number of groups never have to talk to each other, they only communicate internally. So rather than use load balance and Redis to achieve large amount of connection, we just have multiple box running and assign group to one of box or move them to another box on fly.

Some of groups only have few hundreds connections, if you split them to multiple box and use Redis that's a bit over kill. Always assign them to one box and have in memory broadcast that's something I think it's proper for this scenario.

Something like slack that you can never talk to another domain, each domain means separate concern. Or similar to IRC, even though some of server can have multiple box but you can have just single box and switch them on wish.

Nness:
I will include tests in future pull request. :)

PR #1033, opened 2016-05-19
QueueNames<T> does not respect QueuePrefix

kinstephen:
Here are the changes again. For some reason it's still showing spacing differences which are coming from the v4 version. If this isn't acceptable I can manually put the changes in if needed.

### DIFF
--- a/src/ServiceStack.Interfaces/Messaging/QueueNames.cs
+++ b/src/ServiceStack.Interfaces/Messaging/QueueNames.cs
@@ -13,15 +13,15 @@ namespace ServiceStack.Messaging
 		{
 			var utf8 = new UTF8Encoding(false);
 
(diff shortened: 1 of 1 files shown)


mythz:
@kinstephen Can you have a look at your commits in this PR? They're rewriting every line ruining the provenance of each line. Commits should only contain the lines that are changed. Maybe you need to start with a clean fork and apply the changes manually then resubmit the PR.

kinstephen:
This is because of the spaces differences to those lines originated from the v4 version. I will manually add the changes this time.

mythz:
You shouldn't be copying code from v4 source files directly. Only fixes.

kinstephen:
Demis, I wasn't. The fix was in v4. I was pulling the fix back from v4. No new code at all.

PR #1034, opened 2016-05-19
Optionally return referenced types when IncludeTypes parameter set in ServiceStack Reference

donaldgray:
This is based on original question on the forum: https://forums.servicestack.net/t/intercepting-extending-nativetypesservice/2419

The approach that I took in that forum post was quite different as I was trying to get this functionality from the outside, rather than making a change internally. 

The issue that I had was that when specifying `IncludeTypes` when using the "Add Service Reference" functionality, it would generate those types and those types only.

I require that when I specify an `IncludeTypes` it will generate it for that class and reference classes. I added another parameter option when generating a service reference called `IncludeRequestReferenceTypes`. If this is `False` (default) then `IncludeTypes` works as it currently does. However, if `True` then data for the `IncludeTypes` and their return types and referenced types will be generated.

E.g. If I have the following:
```csharp
public class NativeTypesTestService : Service {
    public object Any(Dto request){
        return "just a test";
    }
}

public class Dto : IReturn<DtoResponse>{
    public EmbeddedResponse ReferencedType { get; set; }
}

public class DtoResponse  {
    public EmbeddedRequest ReferencedType { get; set; }
}

public class EmbeddedResponse { }
public class EmbeddedRequest { }
```

Currently if I generate this with `?IncludeTypes=Dto` I will only get the `Dto` class generated. However if I run this with `?IncludeTypes=Dto&IncludeRequestReferenceTypes=True` it will generate samples for `Dto`, `DtoResponse`, `EmbeddedResponse` and `EmbeddedRequest` classes.

### DIFF
--- a/src/ServiceStack.Common/MetadataTypes.cs
+++ b/src/ServiceStack.Common/MetadataTypes.cs
@@ -4,6 +4,8 @@ using ServiceStack.DataAnnotations;
 
 namespace ServiceStack
 {
+    using System.Linq;
+
     public class MetadataTypesConfig
     {
         public MetadataTypesConfig(
@@ -27,7 +29,8 @@ namespace ServiceStack
             bool makePropertiesOptional = true,
             bool makeDataContractsExtensible = false,
             bool initializeCollections = true,
-            int? addImplicitVersion = null)
+            int? addImplicitVersion = null,
+            bool includeRequestReferenceTypes = false)
         {
             BaseUrl = baseUrl;
             MakePartial = makePartial;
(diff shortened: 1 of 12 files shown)


mythz:
Hi thanks for the PR, rather than adding a new option that changes the behavior of IncludeTypes can you change it so a `.*` suffix contains the behavior you want, e.g Allow Users to specify `?IncludeTypes=Dto.*` instead of `?IncludeTypes=Dto&IncludeRequestReferenceTypes=True` it makes it easier to use/remember and we're not introducing new concepts that changes behavior of existing features.

donaldgray:
Hi, yes that makes sense. Agree it'll be easier as `.*` will be familiar syntax. Will update PR when I've made the changes.

donaldgray:
@mythz - I've updated to use wildcard suffix rather than a separate param.

mythz:
Sweet, this is an awesome feature thanks!

mythz:
FYI this change is now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

PR #1035, opened 2016-05-19
QueueNames<T> does not respect QueuePrefix (Take 3)

kinstephen:
My apologies Demis. This should be the last time. I've manually gone in and added the changes.


mythz:
This is even worse? There should just be 1 commit with only the lines that were changed.

You're best approach is to start from a clean fork/branch and apply the changes manually.

kinstephen:
Geez sorry. I'm not used to GitHub. I mostly use subversion.

PR #1036, opened 2016-05-19
QueueNames<T> does not respect QueuePrefix (Clean Commit and whitespace)

kinstephen:
Last time? I hope so.


mythz:
perfect, thx

PR #1037, opened 2016-05-25
Move IDbProfiler interface from ServiceStack to ServiceStack.Common assembly

shift-evgeny:
This allows the interface to be implemented in applications not running in the ServiceStack environment, but using ServiceStack.OrmLite.

This is an issue that has come up for others in the past, e.g. http://stackoverflow.com/questions/11530943/servicestack-ormlite-sql-query-logging

It's a breaking change, but should be an easy one to fix for the few people that implemented a profiler.

### DIFF
--- a/src/ServiceStack.Common/Data/ExecuteType.cs
+++ b/src/ServiceStack.Common/Data/ExecuteType.cs
@@ -1,4 +1,4 @@
-﻿namespace ServiceStack.MiniProfiler.Data
+﻿namespace ServiceStack.Data
 {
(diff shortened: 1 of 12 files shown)


mythz:
You didn't need to change the namespace which is the main breaking change, can you add it in a MiniProfiler folder in SS.Common keeping the same namespace?

shift-evgeny:
Oh! Sure, here you go.

mythz:
Cheers :+1:

PR #1039, opened 2016-05-26
Added public IHasDbTransaction interface similar to IHasDbConnection

shift-evgeny:
It allows getting the real transaction from a ProfiledDbTransaction in client code without Reflection hacks.

If this is OK I'll send a PR in ServiceStack.OrmLite to implement this interface in OrmLiteConnection and OrmLiteTransaction, too.

### DIFF
--- a/src/ServiceStack.Common/Data/IHasDbConnection.cs
+++ b/src/ServiceStack.Common/Data/IHasDbConnection.cs
@@ -15,5 +15,10 @@ namespace ServiceStack.Data
     {
         IDbCommand DbCommand { get; }
     }
+
+    public interface IHasDbTransaction
+    {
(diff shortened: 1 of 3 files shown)


mythz:
Hi, can you change this to not use any C# 6.0 features, it breaks the CI build and I end up having to revert back to using C# 5.0 manually.

shift-evgeny:
OK, here you go.

Maybe it's worth setting the language version in the project settings to avoid accidentally using C# 6 features?

mythz:
Cheers.

Don't want to do this for all projects. after August 1st we plan on upgrading all projects to .NET 4.5 after which we'll be able to use C# 6.

PR #1041, opened 2016-06-15
Updated swagger to spec v1.2

carlin-q-scott:
v1.1 isn't supported by swagger-codegen from what I could tell. This upgrade isn't complete but I think it's good enough as it definitely serves my purposes


mythz:
Appreciate the PR but it's too big and disruptive to accept for v3 which doesn't have anyone supporting it, which it would need for any incomplete feature to handle any maintenance requests.

carlin-q-scott:
I understand. As long as others with my same predicament can find the code that works.

PR #1044, opened 2016-06-25
Abstract missing type generator for AutoQueryFeature

bmmathe:
Added an interface for the missing type generator in AutoQueryFeature.
This will give the user more control of the types that are generated for
the auto query feature.  My team needed a way to add custom attributes
to the generated types (specifically Authenticate).  Adding an interface
and base implementation with a hook to add custom attributes allows us
to decorate our request DTOs and override the AddCustomAttributes method
on the base IAutoQueryMissingTypeGenerator implementation.

### DIFF
--- a/src/ServiceStack.Server/AutoQueryFeature.cs
+++ b/src/ServiceStack.Server/AutoQueryFeature.cs
@@ -188,64 +188,17 @@ namespace ServiceStack
         {
             var scannedTypes = LoadFromAssemblies.SelectMany(x => x.GetTypes());
 
-            var misingRequestTypes = scannedTypes
+            var missingRequestTypes = scannedTypes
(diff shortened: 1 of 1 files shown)


bmmathe:
There might be a better alternative solution to accomplish what my team needed but I couldn't find one.  If this is not the best implementation to accomplish what we needed in the comments of the PR please provide a better solution and I will work down that path.

mythz:
The way we allow customization where possible is to use delegate filters where possible, which I've added in this commit: https://github.com/ServiceStack/ServiceStack/commit/9c5b207c6718b1d485c006b136ae56f70b990a4b

Which should let you add further customization for each AutoQuery Service that's generated.

```csharp
Plugins.Add(new AutoQueryFeature {
    GenerateServiceFilter = (typeBuilder, method, requestType) => {
    }
});
```

PR #1045, opened 2016-06-27
Fix exception when trying to get member from base type

xplicit:
Fixes https://github.com/ServiceStack/Issues/issues/449


mythz:
Cheers 👍

PR #1047, opened 2016-07-14
TypeScript generation fixes

Layoric:
Fixes for TypeScript generation:
- Order of base types causing issues if base class is declared after for TS 1.8
https://github.com/Microsoft/TypeScript/issues/4341
- `createResponse` generating invalid code for primitives, eg `return new string()`.


mythz:
Cheers :+1:

PR #1048, opened 2016-07-14
Handle disconnected ServerSentEvent subscriptions

TheDruidsKeeper:
Currently if a client disconnects while a channel is being notified, it causes an error to be thrown when attempting to SetResult on the TaskCompletionSource. This patch will prevent setting the result on the
task if it's already completed. It will also prevent attempting to notify a subscription where the underlying connection is already closed (currently the notify calls publish which throws an exception which then
unsubscribes the current subscription - this will prevent the exception, and simply unsubscribe.)


TheDruidsKeeper:
Test failures appear to be unrelated to this pull request as they were already failing due to previous commits.

mythz:
Cool, yeah these are just some of the temporal integration tests than can fail with our build server, ran it again and all's good. Cheers 👍

PR #1053, opened 2016-08-25
MsgPack minor changes to support .net pcl/core/standard

christallire:
Modified usage of deprecated reflection implementation to support both .net framework and .net pcl/core/standard


mythz:
Hi thanks for the PR, although we've got [our own PCL-compatible wrappers](https://github.com/ServiceStack/ServiceStack.Text/blob/f455d0bf6c3249337a770363f847f64067f3c3f0/src/ServiceStack.Text/ReflectionExtensions.cs) which we use to impl pcl-compatible APIs which I've modified in this commit: https://github.com/ServiceStack/ServiceStack/commit/6daa602a38127e2c3bb44fad673be73ac607af2d

PR #1054, opened 2016-08-30
jQuery 3.x Compatibility

TheDruidsKeeper:
Changed $(window).unload() to $(window).on("unload", ) for compatibility with jQuery 3.x (they've removed methods deprecated since 1.8): https://github.com/jquery/jquery/issues/2286


mythz:
Great, thx 👍

PR #1055, opened 2016-09-01
ElmahLogFactory.cs: HttpApplication is optional in case of self hosting

RobSchoenaker:
Support for Elmah logging outside of HttpApplication - Self Hosted.

### DIFF
--- a/src/ServiceStack.Logging.Elmah/ElmahLogFactory.cs
+++ b/src/ServiceStack.Logging.Elmah/ElmahLogFactory.cs
@@ -1,49 +1,85 @@
 using ServiceStack.Logging;
 using System;
 using System.Web;
+using Elmah;
 
 namespace ServiceStack.Logging.Elmah
 {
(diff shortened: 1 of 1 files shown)


mythz:
Hi Rob, can you [approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so we can merge this, thx.

I'm assuming from this PR, it still has the same behavior in ASP.NET when passing in the HttpApplication as before?

RobSchoenaker:
Approved the CLA.

Original behavior is the same, yes. It's 'only useful if you do not have a HttpApplication' like with self hosting - or if you want to pass a different Httppalliction instance (not sure why you would want that, but it's possible)

mythz:
Great, thx 👍

PR #1058, opened 2016-09-02
Add SocketExtensions

xplicit:


mythz:
Why is this PR so big, it should just contain the changes?

mythz:
I've hand picked SocketExtensions and added it in this commit: https://github.com/ServiceStack/ServiceStack/commit/41f3dbb64f36eb2f46af751437430f5d5fd7d464

I've just checked in project.json changes to ServiceStack netcore projs, can you do a clean update to master so we can avoid big PR's like this.

PR #1059, opened 2016-09-08
Preserve port from loadbalancer/proxy in front of ServiceStack

jhf:
When ServiceStack is running behind a loadbalancer or proxy that correctly sets the header
X-Forwarded-Port, then the base url should change to reflect the proxy.
This allows a setup where ServiceStack listens to a specific port, but that port is only exposed to the
loadbalancer/proxy, as in the scenario:
http://my.domain/ -> LoadBalancer -> http://a.b.c.d:x/
where access to, for instance http://my.domain/ should have working links.

Note:
* I tried to do early return without any parsing for any non relevant cases.
* There is currently a transformation Uri->string->Uri->StringBuilder->Uri that I'm not happy with,
  suggestions for improvement that doesn't require logic duplication is welcome.
* I didn't benchmark (or test) this yet - I assume a testcase is in order - any suggestions to an appropriate placement in this project is also welcome.

### DIFF
--- a/src/ServiceStack/HttpRequestExtensions.cs
+++ b/src/ServiceStack/HttpRequestExtensions.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Net;
 using System.Net.Sockets;
 using System.Runtime.Serialization;
+using System.Text.RegularExpressions;
 using System.Web;
 using System.Web.Hosting;
 using ServiceStack.Data;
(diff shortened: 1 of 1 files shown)


mythz:
This impl is too disruptive to accept for a feature that's never been requested, I'm also concerned that this behavior is an unwanted breaking change. So instead of baking it in the framework I've moved the impl of `GetBaseUrl()` to AppHost where it's now overridable so you can rewrite the URLs in your AppHost with:

```csharp
public override string GetBaseUrl(IRequest httpReq)
{
    var baseUrl = base.GetBaseUrl(httpReq);
    return HandleProxyPort(baseUrl, UseHttps(httpReq), ProxyPort(httpReq));
}
```

This change is available from v4.5.1 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

I'll revisit this feature if we get more requests for it in future, but we'd need a more maintainable and configurable way to opt into using URL filters.

jhf:
Thank you for adding the possibility of a hook.
While I understand that you did not take the implementation, I was expecting that you would think such a feature sensible, mostly because of commit: ab363d931c810272c01279d671a673116c5f6822 , that, in my mind, does the same kind of thing, but for https instead of the port.
One possibility would be a plugin that honors all of the generally available "X-*" headers a sensible proxy might set.
Currently http(s) and client ip provided as headers by a proxy are preserved automatically based on the headers, while the port is not, hence my attempt at the patch, which seems like an inconsistency.

Anyway, thank you for your good work!

jhf:
FYI @mythz:
I'm now on the second installation where this is required - and where it is surprising that ServiceStack does not automatically use the port of the loadbalancer, as exposed by the http header

mythz:
This feature has never been requested before and has the effect of changing default behavior where it's not clear it's desired (and can potentially break existing systems not wanting it). Anything that changes default behavior like this needs to be configurable (i.e. opt-in/out). The change I've added makes it easy to opt into, given it's never been requested for this is preferred over baking it into the framework .

This implementation is also untested, fairly disruptive and inefficient, not something we'd be able to accept as-is. You can submit it as a feature request at https://servicestack.uservoice.com so we can tell if there's any demand for it.

PR #1061, opened 2016-09-12
Fix auth session extends on second authentication

xplicit:
This commit makes saving session with SessionExpiry when user authenticated second time while session is active. Possible IAuthProvider should have `SessionExpiry` property to access it without casting to abstract class. 

https://forums.servicestack.net/t/session-never-expires/2906


mythz:
Looks good, thx.

PR #1062, opened 2016-09-13
Add number to primitiveDefaultValues

TheDruidsKeeper:
Added "number" to primitiveDefaultValues because:
1. AppendType calls options.ImplementsFn to get type info for generating the responseTypeExpression (`createResponse() { return x; }`).
2. ImplementsFn calls Type(...) to get the type name of the return object
3. `string Type(string type, string[] genericArgs)` returns from TypeAlias

When generating the type definition for IReturn<int> from the above workflow, the result was the following (invalid) syntax: `createResponse() { return new number(); }`
The result will now be this: `createResponse() { return 0; }`

### DIFF
--- a/src/ServiceStack/NativeTypes/TypeScript/TypeScriptGenerator.cs
+++ b/src/ServiceStack/NativeTypes/TypeScript/TypeScriptGenerator.cs
@@ -68,6 +68,7 @@ namespace ServiceStack.NativeTypes.TypeScript
             {"Single", "0"},
             {"Double", "0"},
             {"Decimal", "0"},
+            {"number", 0},
             {"List", "[]"}
         };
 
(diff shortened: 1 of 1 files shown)


mythz:
Can you provide an example of a DTO that this fixes?

TheDruidsKeeper:
Dto in C#:
```C#
[Route("/game/teams/create")]
[Api("Groups the current pending travelers into a team and tells the game server to send them into The VOID")]
public class CreateTeamRequestDto : IReturn<int>
{
	public int GateId { get; set; }
}
```

Generated typescript:
```javascript
/**
* Groups the current pending travelers into a team and tells the game server to send them into The VOID
*/
// @Route("/game/teams/create")
// @Api("Groups the current pending travelers into a team and tells the game server to send them into The VOID")
export class CreateTeamRequestDto implements IReturn<number>
{
	GateId: number;
	createResponse() { return new number(); }
	getTypeName() { return "CreateTeamRequestDto"; }
}
```

mythz:
Services should return a Reference type, typically a Response DTO but can be a raw value like `string` or `byte[]`. Returning value types like `int` is not supported and wont work in other Add ServiceStack Reference languages.

I'll include merge this, but your Services shouldn't be returning value types.

PR #1067, opened 2016-09-22
add overwriteable method on authProvider for session value persistance.

joelharkes:
Im trying to persist an Cart Id (eCommerce platform) on user log-in.

The requirement is that the user keeps his cart when he logs in.

I save the cartId i the session when the user logs in.

But in my subclass of the CredentialsAuthProvider onAuthenticated the CartId is gone.

### DIFF
--- a/src/ServiceStack/Auth/CredentialsAuthProvider.cs
+++ b/src/ServiceStack/Auth/CredentialsAuthProvider.cs
@@ -93,14 +93,20 @@ namespace ServiceStack.Auth
             return Authenticate(authService, session, userName, password, string.Empty);
         }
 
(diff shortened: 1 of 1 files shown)


mythz:
Hi Joël, this is fine but you can also use the [Session Bag](https://github.com/ServiceStack/ServiceStack/wiki/Sessions#using-dynamic--untyped-sessions-in-servicestack) to store info about a Users Session that you can access after the User has authenticated but you'll need to set `AuthFeature.GenerateNewSessionCookiesOnAuthentication = false` when you register the `AuthFeature` plugin.

joelharkes:
👍 
Thanks for the info.

joelharkes:
yea but it would mean an extra redis call for the session bag. Thanks for the quick reply

PR #1068, opened 2016-09-22
[netcore] System.Threading.ThreadPool package downgrade warning

KevinHoward:
When I run `dotnet restore` on **ServiceStack.Core**, I get the following warning:

_warn : Detected package downgrade: System.Threading.ThreadPool from 4.0.10 to 4.0.1_ 

So, this PR is to bump the version. I didn't investigate whether or not there was a reason for picking 4.0.1, but this was the only dependency throwing the warning, so I assumed it was a typo.

### DIFF
--- a/NuGet.Core/ServiceStack.Core/ServiceStack.Core.nuspec
+++ b/NuGet.Core/ServiceStack.Core/ServiceStack.Core.nuspec
@@ -43,7 +43,7 @@
         <dependency id="System.Runtime.Serialization.Primitives" version="[4.1.1, )" />
         <dependency id="System.Collections.Specialized" version="[4.0.1, )" />
         <dependency id="System.Linq.Queryable" version="[4.0.1, )" />
-        <dependency id="System.Threading.ThreadPool" version="[4.0.1, )" />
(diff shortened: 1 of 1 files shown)


mythz:
Sweet thx for the .NET Core PR Kevin!

KevinHoward:
The .NET Core work is looking good. @xplicit and @mythz keep up the good work!

mythz:
Thx Kevin! BTW not sure if you've noticed it as well, but we've ported some live demos over at: https://github.com/NetCoreApps/ if you wanted to kick the tires early :)

KevinHoward:
Thanks Demis! I'll check them out. 

The reason for my early tinkering was that I currently have two solutions (front-end is .net core SPA, and the api is ServiceStack on v4.5) and I am looking to combine the two into a single .NET Core project. My only hurdle is routing api calls, so the MVC demo looks to be promising.

mythz:
Yeah in .NET Core ServiceStack is just a standard module, if the route isn't handled by ServiceStack it just passes through on to the next module which will now let you use ServiceStack + MVC together in the same route namespace.

PR #1072, opened 2016-09-29
Non-session optimized implementation of wire serializer

wwwlicious:
Adding new format "wire". a fast Binary serializer for POCO objects
https://github.com/AsynkronIT/Wire

I am using the most tolerant Versioned mode which from the docs says

"Versioned, in this mode, Wire will emit both type names and field information in the output stream. This allows systems to have slightly different versions of the contract types where some fields may have been added or removed."

In this mode is isn't required to register the types for serializing in advance as it will as stated above emit the type and field info to the output stream so AFAIK you do not need to hold a dictionary of types as MsgPackFormat does.

I am also not sure of the lifecycle of serialization sessions so am using a single static instance of the serializer in the client.

### DIFF
--- a/src/ServiceStack.Wire/Properties/AssemblyInfo.cs
+++ b/src/ServiceStack.Wire/Properties/AssemblyInfo.cs
@@ -1,0 +1,25 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("ServiceStack")]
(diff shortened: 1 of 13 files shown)


wwwlicious:
Build is failing due to the use of MimeType.Wire which requires ServiceStack.Text reference to be updated in lib folder. I did this manually locally but didn't want to commit in case it has specific build reqs

mythz:
Great thx Scott, much appreciated 💯 I can update to the latest SS.Text dll on my side.

wwwlicious:
no probs

mythz:
FYI, ServiceStack.Wire NuGet package is [now on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

wwwlicious:
sweet! you sir, are a prince among men :+1:

PR #1074, opened 2016-10-02
Release notes outdated -> linked correctly

scottmcarthur:
The release notes shown under the code root directory are from 2014, this makes the project look outdated, unless you are will to dig and find the notes that are actively being maintained. This change links to the latest and gives access to older notes too.


mythz:
no worries thx Scott.

PR #1075, opened 2016-10-07
Added generic and string ThrowIfNull so that it can be used as part of an inline…

stuartbfs:
… assignment rather than having to take 2 lines.  This just allows a null check to take place on the same line as an assignment operator.

E.g.

```C#
class SomeClass
{
   private readonly string internalValue;

   public SomeClass(string externalValue)
   {
       this.internalValue = externalValue.ThrowIfNullOrEmpty(nameof(externalValue));
      
      // instead of:
      // externalValue.ThrowIfNullOrEmpty(nameof(externalValue));
      // this.internalValue = externalValue;
   }
}
```

### DIFF
--- a/src/ServiceStack.Common/AssertExtensions.cs
+++ b/src/ServiceStack.Common/AssertExtensions.cs
@@ -25,23 +25,35 @@ namespace ServiceStack
                 throw new ArgumentNullException(varName ?? "object");
         }
 
-        public static void ThrowIfNullOrEmpty(this string strValue)
+        public static T ThrowIfNull<T>(this T obj, string varName)
         {
-            ThrowIfNullOrEmpty(strValue, null);
+            if (obj == null)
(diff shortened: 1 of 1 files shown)


stuartbfs:
Seems to break due to something in the ServiceStack.Aws.DynamoDb.DynamoDbAuthRepository repository not being able to find the string ThrowIfNullOrEmpty after adding return value to it - not sure if it maybe needs to pull in the updated dependency?

mythz:
No worries, I'll check it out, thx!

mythz:
Yep just needed to update the deps, this change is available from v4.5.1 that's [now on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet) if you need it.

stuartbfs:
Great thanks very much!

PR #1077, opened 2016-10-11
Make the container behaviour in AppHost extensible

joelharkes:
I would like to make the container getter extensible in the appHost.

I was overwriting all the resolve, tryResolve, register, etc methods to add to the child container after startup. But Sometimes some code references to the AppHost.Container directly. Since this method is public why not make it overrideable. then with one override i can solve all my problems.

Usecase: Supply a childContainer (to plugins) per request or per customer  (depending on performance/memory usage)


joelharkes:
So i'd suggest either making the container protected or making the container overridable. i did the second option so it would be backwards compatible

mythz:
Yep no worries, we're all for making things overridable when needed.

PR #1078, opened 2016-10-13
Making ServiceController extensible.

joelharkes:
Since `IAppHost` directly references to `ServiceController` instead of `IServiceController` for whatever reason I do not know? (Either extend the interface or remove it completly imo).

Usecase:
It like to add some fallback strategy for when Service cannot be found.

Btw i cant build the project because pfx key file is missing, and when i remove it it dies on DynamoDbv2/CodeAnalysis.dll is missing  (nuget package version wanted 2.3.1.5.3, though version 2.3.3.0 was installed)

### DIFF
--- a/src/ServiceStack/Host/ServiceController.cs
+++ b/src/ServiceStack/Host/ServiceController.cs
@@ -505,7 +505,7 @@ namespace ServiceStack.Host
             return Execute(requestDto, new BasicRequest());
         }
 
-        public object Execute(object requestDto, IRequest req)
+        public virtual object Execute(object requestDto, IRequest req)
         {
             req.Dto = requestDto;
             var requestType = requestDto.GetType();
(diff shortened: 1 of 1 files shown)


joelharkes:
It would be great if i could add Assemblies to reflected (to discover services) but i guess this is also not possible after Init()?

mythz:
Yeah I'll remove `IServiceController` at some stage was used initially to expose a limited user-facing API.

It's not supported after Init() for same reasons, on StartUp it can be registered dynamically using RegisterService APIs.

joelharkes:
to be honest i think you should keep the IServiceController and just extend it to have to proper public functions. So it stays more like a strategy pattern. 

Probably do the same thing for the ioc container. The ContainerAdapter interface is nice but i dont think it sustains anymore (seeing ss.funq.container uses named registrations which is not in the interface). 

Make an (as simple as possible) ioc container interface, which SS itself can work with. So people can completly bring there own ioc container using this interface. the advantage would be you would be able to register your implementations through this interface as well, creating a true abstraction (instead of only abstracting the resolving part).

mythz:
It's effectively an internal interface with a single implementation and is now just a source of friction. Any extension points can be handled by overriding existing virtual methods on `IAppHost`, adding any additional ones as they're needed.

From next release In .NET Core, you'll be able to plug into your preferred IOC in .NET Core's `StartUp` class and the Container Adapter will resolve dependencies from .NET Core's abstraction.

PR #1079, opened 2016-10-13
Fix build error referencing wrong dll version dynamo.dll

joelharkes:
when i restore packages and try to build it throws an error it cant find this dll.

apparently it downloaded another version of the dll.


joelharkes:
After removing the pfx file reference. (did not commit this ofc)

it builds again:
image

joelharkes:
I wonder why it builds in the first place? maybe the package folder does not get cleaned up on build?

mythz:
hmm looks like a dirty NuGet upgrade, thx. Looks like Analyzers have different behavior than dlls.

PR #1080, opened 2016-10-14
Apply documentation for Apphost and RequestContext.

joelharkes:
Additional fixes
* Fix typo DispsableTracker => DisposableTracker
* Move disposableTracker class to its own file.
* Fix use resolve instead of tryResolve so right exception is thrown.


joelharkes:
i think i was working on a wrong branch im making a proper merge right now

mythz:
ok np

joelharkes:
not done yet

mythz:
BTW we need to run [codegen-projs.js](https://github.com/ServiceStack/ServiceStack/blob/master/build/codegen-projs.js) every time we add/delete files in order to resync + regenerate the .Signed / PCL projects.

I use this custom external tool to run any node script, just open the script in VS.NET and hit `Alt T + Y`

image

joelharkes:
I see,
im trying to run it directly through node but that throws errors?
image

mythz:
oh, yeah it's expected you have cloned all other SS Repos. That's ok, I'll run it and check the generated projs after you're done.

joelharkes:
woudln't it be a better idea to implement this either inside the teamcity build agent (probably not optimal because you cant open it manualy). or Run a gulp watcher on .proj files?

joelharkes:
Also a good thing to note is that the whole node_modules (only 1 folder though) is included in git.

image

mythz:
I still need to configure which projs need which projs generated, this is 1 script for all projs in all repos which takes me seconds using R# to find file + key shortcut to run. I prefer it static w/o needing CI as I can run and open the solution locally without waiting for CI to run (takes a long time) and check it in, then pull the changes, etc.

The new project.json system MS will move to should save all this effort which natively support wildcard files globs so wont need to maintain an individual list.

Checking in node_modules here is ok it keeps the script self-contained without needing to install npm packages.

joelharkes:
Will this be sufficient? adding it to signed.csproj. 

Yea microsoft is making it easier, as well as for managing package version dependencies.

mythz:
Yep that's all the script does

joelharkes:
mm here it builds manualy dunno why TC doesn't pick it up.

also i always have to remove servicestack-sn.pfx from servicestack.interfaces.csproj to make it build. is there some way to make it work? i could just input a fake private key?

mythz:
We've removed ServiceStack.Interfaces.csproj from the build configuration, all projects should be referencing `/lib/ServiceStack.Interfaces.dll` instead.

joelharkes:
hey mythz whats the build error why doesnt it merge?

also what if i wanted to add an interface or update and interface in servicestack.Interfaces. how would i make this work? do i copy the dll manually?

mythz:
It's just a flaky test, it builds now.

Only we can build and sign the ServiceStack.Interfaces .csproj, we then put the signed .dll in `/lib` which is what all proj references.

mythz:
Let me know when you've finished with this PR and are ok to merge.

joelharkes:
Yea can be merged. (With not finished I meant that the PR had corrupt/wrong changes). When I'll continue/add more I'll make a new PR.

mythz:
ok cool, thx!

PR #1096, opened 2016-11-18
Swagger Bootstrap: Show required request DTO properties

schmidt4brains:
Added a "Required" column to the Parameters table.
Removed the "Required" column from the Response Model table.

The logic was simply in the wrong place.

Before the change:
image

After the change:
image


mythz:
Great, thx Doug 👍

schmidt4brains:
Ah, good. I was trying to see why the TeamCity pull-request build was failing, and couldn't make see any related cause.

PR #1097, opened 2016-11-21
Updated OrmLiteCacheClient for SQL Server in-mem optimized tables

KevinHoward:
Updated OrmLiteCacheClient for SQL Server in-memory optimized tables but interfacing out the CacheEntry, plus restricted the SqlServerMemoryOptimizedAttribute to decorate classes only.


KevinHoward:
Hey guys! 

My guess is that the CI server may be failing because either the edition/version of SQL Server doesn't support In-Memory Optimized tables (2014 [**not express**] or 2016+ any version) or the _test_ database is not set up with a memory optimized filegroup ([https://msdn.microsoft.com/en-us/library/dn639109(v=sql.120).aspx](https://msdn.microsoft.com/en-us/library/dn639109(v=sql.120).aspx)).  

Here is a link to a good write up on this topic along with requirements and limitations between server versions [https://www.simple-talk.com/sql/learn-sql-server/introducing-sql-server-in-memory-oltp/](https://www.simple-talk.com/sql/learn-sql-server/introducing-sql-server-in-memory-oltp/). 

I am using 2016 developer, but should probably install a copy of 2014 and do some further testing, particularly around the collation and indexing. Looks like I may need to add a property-level attribute to handle the indexing and bucket count.

mythz:
ok cool, we'll likely need to make them manual tests until we setup the CI Server properly.

PR #1098, opened 2016-11-22
Add SQL Server Collate and BucketCount attributes

KevinHoward:


mythz:
Thx Kevin, FYI I've just checked in the updated SS.Interfaces .dll to ServiceStack + OrmLite repos.

KevinHoward:
Great! I'll start doing some more testing. I may also take a  look at System-Versioning Temporal Tables ([https://msdn.microsoft.com/en-us/library/dn935015.aspx](https://msdn.microsoft.com/en-us/library/dn935015.aspx)) for the 2016 provider too.

PR #1099, opened 2016-11-29
SSE - Skip broadcasting to expired subscriptions

TheDruidsKeeper:
If the subscription is expired or closed, then it shouldn't be receiving publications.


mythz:
Agreed! thx for the PR 👍

PR #1100, opened 2016-11-30
Updated ContentFormat.IsBinary() logic for Wire format

donaldgray:
Added Wire MimeType to IsBinary() check as I was getting an exception when trying to cache Wire formatted content without this.

I also noticed there currently isn't support for Wire format in `[Restrict(RequestAttributes.Wire)]`. I had a shot at doing this locally but it meant touching ServiceStack.Interfaces project (to update RequestAttributes and Feature classes) which resulted in a few issues relating to needing to update ServiceStack.Interfaces for other projects while I'm unable to sign it.

If Restrictions for Wire is something you think should be supported I can add the changes I've made to this P.R. (albeit they may be incomplete).

### DIFF
--- a/src/ServiceStack.Client/ContentFormat.cs
+++ b/src/ServiceStack.Client/ContentFormat.cs
@@ -69,6 +69,7 @@ namespace ServiceStack
                 case MimeTypes.MsgPack:
                 case MimeTypes.Binary:
                 case MimeTypes.Bson:
+                case MimeTypes.Wire:
                     return true;
             }
 


mythz:
FYI I've just added Wire as a first-class format in this commit: https://github.com/ServiceStack/ServiceStack/commit/03dc5bc442a358c91abdcff9e1263936a183bbb6 - I've relegated Yaml to Other since it's not a built-in format.

The ServiceStack.Interfaces.dll has been updated and this change is now available from v4.5.5 that's [now on MyGet](http://docs.servicestack.net/myget).

donaldgray:
Nice one, thanks Demis

PR #1102, opened 2016-12-21
Performance improvements

xplicit:
.NET Core benchmark  `/json/reply/hello?name=world` increased performance from 38K rps to 46K rps on desktop Core I5 (4 cores) and from 26K rps up to 29K rps on Azure Standard_F4s machine


mythz:
Great thx Sergey, the NameValueCollection was one of the areas I wanted to better integrate with .NET Core's implementation, great to see it yielding nice perf improvements.

PR #1108, opened 2017-01-25
Fix InvalidCastException when redirect after wrong auth

xplicit:


joelharkes:
Ah yes!!

We have been waiting for this. i had manually copied a whole bunch of code to a subclass of Authenticate Attribute. because a lot of code was static & non virtual.

PR #1109, opened 2017-01-25
Update SocialExtensions.cs

glaidler:
use https for gravatar so usage within html on https site won't give browser warnings.


mythz:
Cheers 👍

PR #1113, opened 2017-02-06
Minor Metadata Fixes

englerj:
- Restricted visibility of the Request Logs service to none. 
- Fixed a minor issue where the Request Logs debug link would not work when debugging on a local IIS server.

### DIFF
--- a/src/ServiceStack/RequestLogsFeature.cs
+++ b/src/ServiceStack/RequestLogsFeature.cs
@@ -96,7 +96,7 @@ namespace ServiceStack
             }
 
             appHost.GetPlugin<MetadataFeature>()
-                .AddDebugLink(AtRestPath, "Request Logs");
+                .AddDebugLink(AtRestPath.TrimStart('/'), "Request Logs");
         }
     }
 }
--- a/src/ServiceStack/Admin/RequestLogsService.cs
+++ b/src/ServiceStack/Admin/RequestLogsService.cs
@@ -48,6 +48,7 @@ namespace ServiceStack.Admin
     }
 
     [DefaultRequest(typeof(RequestLogs))]
+    [Restrict(VisibilityTo = RequestAttributes.None)]
     public class RequestLogsService : Service
     {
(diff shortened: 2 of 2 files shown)


src/ServiceStack/RequestLogsFeature.cs
```diff
@@ -96,7 +96,7 @@ public void Register(IAppHost appHost)
             }
 
             appHost.GetPlugin<MetadataFeature>()
-                .AddDebugLink(AtRestPath, "Request Logs");
+                .AddDebugLink(AtRestPath.TrimStart('/'), "Request Logs");
         }
```
mythz: The issue should be fixed in `.AddDebugLink()` or when link is displayed, but not here. 
What's the url that gets displayed?

englerj: I think I should have added just a bit more to that comment. This issue is only present when debugging on a local IIS server and the website is under a virtual directory.
In my case, I have a DemoApp project with a URL of _http://localhost/DemoApp/_.
The metadata and everything else comes up correctly at _http://localhost/DemoApp/*_
However, the request logs link is _http://localhost/requestlogs_ when it should be _http://localhost/DemoApp/requestlogs_

englerj: I think I should have added just a bit more to that comment. This issue is only present when debugging on a local IIS server and the website is under a virtual directory.
In my case, I have a DemoApp project with a URL of _http://localhost/DemoApp/_.
The metadata and everything else comes up correctly at _http://localhost/DemoApp/*_
However, the request logs link is _http://localhost/requestlogs_ when it should be _http://localhost/DemoApp/requestlogs_

mythz:
I've added a complete fix for the debug/plugin links so it works in all different hosting scenarios in this commit: https://github.com/ServiceStack/ServiceStack/commit/c17323cc532239daff432d43cfa0e85d5b3e9920

englerj:
That fixes the issue I was having. Thanks for the update!

PR #1114, opened 2017-02-12
Fix session Cookie creation to respect HostContext.Config.RestrictAll…

zigmo:
…CookiesToDomain in .NET Core

Removed `NetCoreCookies` class to use the explicit implementation in `NetCoreResponse.SetCookie()` as it's already use bits for .NET Core in [ToCookieOptions()](https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack/Host/Cookies.cs#L157)

### DIFF
--- a/src/ServiceStack/Host/Cookies.cs
+++ b/src/ServiceStack/Host/Cookies.cs
@@ -15,12 +15,6 @@ namespace ServiceStack.Host
 
         public static Cookies CreateCookies(IHttpResponse httpRes)
         {
-#if NETSTANDARD1_6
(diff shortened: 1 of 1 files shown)


xplicit:
Can you make this PR to `master` branch? `netcore` is development branch and changes from them will be merged later to master, but this fix is preferably to have in master before other changes.

To do it you can do the following commands:

    git checkout master
    git pull origin master
    git cherry-pick <sha-of-the-cookies-commit>
    git push origin master

and then make PR again based on master branch.

PR #1124, opened 2017-03-10
Send request on JsonHttpClient.Patch(IReturnVoid)

svantreeck:
This makes the JsonHttpClient mirror the behavior in [ServiceClientBase](https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack.Client/ServiceClientBase.cs#L1444) that passes the request object to the server.


mythz:
Cheers 👍

mythz:
FYI this is now available from v4.5.7 that's now [available on MyGet](http://docs.servicestack.net/myget).

PR #1126, opened 2017-03-17
Fix RequestContext going out of scope in .NetCore

joelharkes:
Hey mythz, i posed some weird stackoverflow issue lately of getting weird redis errors.

I traced the bug to in our code but it i found the issue also lays in servicestack core. 
The requestContext can easily be turned into a static dictionary when touched directy & indirectly in startup. Or outside of a request. This would already create the AsyncLocal<Dictionary> and thus this dictionary would be given anywhere deeper in the pipeline: also every request gets this same Dictionary.

I solved it this way it now overrides this dictionary at the beginning over every request.

Another option, which wouldn't be water-thight but maybe a good addition: on `CreateItems` check if `ServiceStackHost.Instance.HasStarted` Otherwise throw an error, so people will see there bugs in code early on instead of getting weird errors when concurrent calls start happening.


mythz:
Looks good, thx Joel 👍

PR #1130, opened 2017-03-24
Adds CacheClientPrefixer that decorates ICacheClient commands with prefix

joelharkes:
Adds ability to prefix the cache client using a decorator.

for example can be added on RedisManagerPool => ClientFactory for easy multi tenant use of the redis instance.

### DIFF
--- a/src/ServiceStack/Caching/CacheClientPrefixer.cs
+++ b/src/ServiceStack/Caching/CacheClientPrefixer.cs
@@ -1,0 +1,134 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ServiceStack.Caching
+{
+    /// <summary>
+    /// Decorates the ICacheClient (and its sibblings) prefixing every key with the given prefix
+    /// 
+    /// Usefull for multi-tenant environments
+    /// </summary>
+    public class CacheClientPrefixer : ICacheClient, ICacheClientExtended, IRemoveByPattern
+    {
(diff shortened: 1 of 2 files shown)


mythz:
Note we don't use `_` prefixes in the code-base, also not a fan of the name which stands out, we don't really Verbify class names, they're usually nouns. I'll let you know after I make some updates.

mythz:
FYI renamed it to `CacheClientWithPrefix` in [this commit](https://github.com/ServiceStack/ServiceStack/commit/539e01352a8cef2a17657d9be257a711a9d2fa4c). 

I've also added an extension method which is likely its preferred usage, e.g:

```csharp
container.Register(c => 
    c.Resolve<IRedisClientsManager>().GetCacheClient().WithPrefix("myprefix"));
```

joelharkes:
Ah i see sorry for the code style. I normaly work with default microsoft code style. is there any resharper file i could import for the complete right code style? 
 
I Like the extension method but the classname is very confusing since it only wrappers/decorates the cacheclient and is not a chache client in itself. But thanks for cleaning up, ill keep these code styles next commit.

mythz:
The name should ideally reflect what it does to aid in discovery/usage not how it's implemented. 

I while ago I published my R# settings at [resharper-settings.DotSettings](https://github.com/ServiceStack/ServiceStack.Aws/blob/master/build/resharper-settings.DotSettings)

PR #1133, opened 2017-03-27
Added a builder class for MultiAppSettings.

englerj:


mythz:
Great thx for the PR 👍 

Please note we avoid usage of `_` prefixes in the code-base (I'll make the updates here).

PR #1137, opened 2017-03-28
Update MockHttpResponse to give clue in testing

jezzsantos:
When testing an using the `MockHttpRequest`, if you forget to initialise an AppHost in your tests (usually done once for all tests) you get the dreaded 'Object reference not set to an instance of an object.' NOT from `MockHttpRequest` but from `MockHttpResponse` with no clue as to the cause. 

This change ensures that you get a clue, that prompts you to initialise a ServiceStackHost in your tests first. It could even go further and point you to a page to demonstarte how to do that.

### DIFF
--- a/src/ServiceStack/Testing/MockHttpResponse.cs
+++ b/src/ServiceStack/Testing/MockHttpResponse.cs
@@ -16,8 +16,7 @@ namespace ServiceStack.Testing
             this.Headers = PclExportClient.Instance.NewNameValueCollection();
             this.OutputStream = new MemoryStream();
             this.TextWritten = new StringBuilder();
-            HostContext.AssertAppHost();
-            this.Cookies = HostContext.AppHost.GetCookies(this);
+            this.Cookies = HostContext.AssertAppHost().GetCookies(this);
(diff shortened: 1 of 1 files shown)


src/ServiceStack/Testing/MockHttpResponse.cs
```diff
@@ -16,6 +16,8 @@ public MockHttpResponse(IRequest request = null)
             this.Headers = PclExportClient.Instance.NewNameValueCollection();
             this.OutputStream = new MemoryStream();
             this.TextWritten = new StringBuilder();
+            if (HostContext.AppHost == null)
+                throw new NullReferenceException("A ServiceStackHost has not been initialised in the test.");
             this.Cookies = HostContext.AppHost.GetCookies(this);
```
mythz: Can you instead just change it to: 

```csharp
this.Cookies = HostContext.AssertAppHost().GetCookies(this);
```

That way we don't need to embed the same error messages everywhere.

jezzsantos:
Yes. I can, but `AssertAppHost` is private and needs to be changed to public then.

mythz:
Yep that works.

PR #1140, opened 2017-04-05
Use proper directory for embedded resources

svantreeck:
Currently it's including `openapi-ui`; however, the services looks for them at `swagger-ui` and this is not able to resolve the ui in .net core apps


mythz:
thanks, this was a forgotten left-over refactor.

PR #1141, opened 2017-04-12
Create generated code with predictable [ApiMember] property ordering.

schmidt4brains:
Generated code for annotated DTOs was creating [ApiMember] with random property ordering. (I suspect the order of the `PropertyInfo` collection is not guaranteed by the CLR.

Each time we regenerated a client library from an endpoint using the built-in `/types/csharp` endpoint on an annotated DTO, the resulting diff was massive, even though the DTO shapes had not changed.

Given an annotated DTO of:
```C#
public class Item
{
    [ApiMember(Description = "The ID of the item", IsRequired = true)]
    public int Id { get; set; }
}
```

One run of the type generator would yield:
```C#
public class Item
{
    [ApiMember(Description="The ID of the item", IsRequired=true)]
    public int Id { get; set; }
}
```

After the next daily build, with no code changes, a subsequent run of the type generator would yield:
```C#
public class Item
{
    [ApiMember(IsRequired=true, Description="The ID of the item")]
    public int Id { get; set; }
}
```

A quick solution was to simply sort the property collection by name to yield a deterministic output from the code generators.

### DIFF
--- a/src/ServiceStack/NativeTypes/NativeTypesMetadata.cs
+++ b/src/ServiceStack/NativeTypes/NativeTypesMetadata.cs
@@ -552,6 +552,7 @@ namespace ServiceStack.NativeTypes
                 .Select(pi => ToProperty(pi, attr))
                 .Where(property => property.Name != "TypeId"
                     && property.Value != null)
+                .OrderBy(property => property.Name)
                 .ToList();
         }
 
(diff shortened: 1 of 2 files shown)


mythz:
ok great, thx Doug 👍

PR #1143, opened 2017-04-17
Check for null ServiceRoutes

mikepugh:
ServiceRoutes is null by default and this was causing an exception to occur during application startup that was swallowed by the AppHost init procedure.


mythz:
ok thanks, tho needs a little whitespace love - running `Ctrl+K,D` in VS.NET should do the trick!

PR #1144, opened 2017-04-18
Issues found by PVS-Studio

skhrenov:
Hello,
I have found a few bugs using PVS-Studio analyzer. PVS-Studio is a static code analyzer for C, C++ and C#: [https://www.viva64.com/en/pvs-studio/](https://www.viva64.com/en/pvs-studio/)

Analyzer warnings:

- [V3095](https://www.viva64.com/en/w/V3095/) The 'feature' object was used before it was verified against null. Check lines: 89, 99. ServiceStack AuthProvider.cs 89

- [V3095](https://www.viva64.com/en/w/V3095/) The 'type.EnumNames' object was used before it was verified against null. Check lines: 286, 294. ServiceStack KotlinGenerator.cs 286

- [V3095](https://www.viva64.com/en/w/V3095/) The 'session' object was used before it was verified against null. Check lines: 122, 170. ServiceStack ServerEventsFeature.cs 122

- [V3042](https://www.viva64.com/en/w/V3042/) Possible NullReferenceException. The '?.' and '.' operators are used for accessing members of the 'session' object ServiceStack ServerEventsFeature.cs 179

In addition, I suggest having a look at the emails, sent from @pvs-studio.com.

Best regards,
Sergey Hrenov,
PVS-Studio Team

### DIFF
--- a/src/ServiceStack/Auth/AuthProvider.cs
+++ b/src/ServiceStack/Auth/AuthProvider.cs
@@ -86,7 +86,7 @@ namespace ServiceStack.Auth
 
             var session = service.GetSession();
             var referrerUrl = request?.Continue
-                ?? (feature.HtmlLogoutRedirect != null ? service.Request.ResolveAbsoluteUrl(feature.HtmlLogoutRedirect) : null)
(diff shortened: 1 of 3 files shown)


mythz:
Thanks but these are mostly false positives which can't happen, e.g. session is always returned and AuthFeature needs to be registered to use AuthProviders.

PR #1152, opened 2017-05-03
InProcessServiceGateway Publish and PublishAsync should now use the verb from the request dto.

englerj:
The original methods would always POST the request. I changed Publish and PublishAsync to mimic what was done in Send and SendAsync where it gets the verb using the IVerb interface of the request.


mythz:
cheers 👍

PR #1155, opened 2017-05-17
Added overloads for register typed filters

englerj:
I added a couple of overloads to ServiceStackHost for convenience purposes.
The overloads allow you to use the container to resolve an ITypedFilter<T> service(s).
It will then delay the resolve until the filter is actually called, allowing the container to correctly auto-wire the dependencies.

`RegisterTypedRequestFilter(c => c.Resolve<MyTypedFilter>());`

Otherwise, the service(s) would be resolved during the initialization, and results in a "Cannot resolve service or one of its auto-wired dependencies" exception. 

`RegisterTypedRequestFilter(container.Resolve<MyTypedFilter>());`

_Note: If there is already a recommended way to do this let me know. I was not aware of any so this is what we did to resolve our problem._

### DIFF
--- a/src/ServiceStack/Host/AppDelegates.cs
+++ b/src/ServiceStack/Host/AppDelegates.cs
@@ -20,6 +20,11 @@ namespace ServiceStack.Host
         void Invoke(IRequest req, IResponse res, object dto);
     }
 
+    public interface ITypedFilter<in T>
+    {
+        void Invoke(IRequest req, IResponse res, T dto);
+    }
+
     public class TypedFilter<T> : ITypedFilter
     {
         private readonly Action<IRequest, IResponse, T> action;
--- a/src/ServiceStack/IAppHost.cs
+++ b/src/ServiceStack/IAppHost.cs
@@ -4,6 +4,7 @@ using System.Net;
 using System.Reflection;
 using System.Threading.Tasks;
 using System.Web;
+using Funq;
 using ServiceStack.Configuration;
 using ServiceStack.Host;
 using ServiceStack.Host.Handlers;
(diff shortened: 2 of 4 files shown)


mythz:
Can you please also provide tests asserting its behavior and showcasing its usage?

englerj:
I added the unit tests in the most recent commit. I was not 100% sure where to put them, as it didn't look like the ServiceStack.Tests project was being used extensively. I put the tests in the same project I saw tests for the other typed filter tests. Let me know if you need that changed.

mythz:
Thanks Josh,

I want to remove `params[]` from the API to remain consistent with other filters, is there any reason why you'd need an array of filters?

englerj:
We are using the filters namely to adjust some of our dtos before they hit the service. Some of these require that we reach out to other services to get some information. Instead of adding that other service as a dependency to our service, we decided to create small filter objects to handle it. Some dtos require that we reach out to multiple services, in which case we would create multiple filters.

mythz:
Is this something you can't do with multiple registrations? Can you update your test with an example?

englerj:
Ah, i misunderstood your previous question.
Can you have multiple typed request filters for a single dto? I was under the impression it was limited to one per dto type, and calling RegisterTypedRequestFilter would replace the current filter.

mythz:
It's registers a single handler per DTO Type indexed by Type which covers most use-cases, in the rarer case when more is needed you can combine them in the same delegate or just register a Global Request Filter and do a type check.

I'd prefer that all message filters contain the same consistent single filter method signature, so unless you need multiple `params[]` I'll remove them. If you do, please update your test to include an example as otherwise it could get refactored away in future.

englerj:
I understand.
These new methods ultimately combine everything in to a delegate (lambda) as you suggested. To keep things consistent, would you prefer I pluralized the names of the new methods, and added proper overloads that do not take a params[] argument? Or would you prefer that the caller always take care of multiple registrations using a delegate?

mythz:
I prefer the caller take care of combining it into a single delegate given it will be rarely needed.

englerj:
ok, I will take care of adjusting what I have to work with that requirement.

PR #1156, opened 2017-05-23
Changed overloads for register typed filters

englerj:
Changed overloads for register typed filters to remove params keyword.

Original: #1155


mythz:
Great thx Josh 👍

PR #1158, opened 2017-07-07
Removed unused locals in RequestBaseWrapper

KevinHoward:
Removed unused local variables from RequestBaseWrapper. 

This should remove the unnecessary null checks against variables that are always null.

### DIFF
--- a/src/ServiceStack/Host/RequestBaseWrapper.cs
+++ b/src/ServiceStack/Host/RequestBaseWrapper.cs
@@ -14,108 +14,69 @@ namespace ServiceStack.Host
         public RequestBaseWrapper(IHttpRequest request)
         {
             this.request = request;
+            Original = (HttpRequestBase)this.request.OriginalRequest;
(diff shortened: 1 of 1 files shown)


src/ServiceStack/Host/RequestBaseWrapper.cs
```diff
@@ -14,108 +14,69 @@ public class RequestBaseWrapper : HttpRequestBase
...
-            get
-            {
-                return contentEncoding ?? Original.ContentEncoding;
-            }
-            set
-            {
-                contentEncoding = value;
-            }
-        }
+        public override System.Text.Encoding ContentEncoding { get; set; }
```
mythz: This is missing the fallback to Original.ContentEncoding

mythz:
looks good, thx Kevin, i'm really liking the new C# 6/7 features myself :)

KevinHoward:
Yes! much cleaner and a lot easier to read. :+1:

PR #1160, opened 2017-07-18
Exclude empty arrays in query string

mikesheen:
Empty arrays pad out the query string length and will cause a 404.15 with ASP.NET projects if the query string exceeds the maximum configured length, this commit excludes empty arrays.

### DIFF
--- a/src/ServiceStack.Client/UrlExtensions.cs
+++ b/src/ServiceStack.Client/UrlExtensions.cs
@@ -481,7 +481,7 @@ namespace ServiceStack
                     continue;
 
                 var value = queryProperty.Value.GetValue(request, true);
-                if (value == null)
+                if (value == null || (value is Array && value.ToJsv() == EmptyArray))
                     continue;
 
                 var qsName = Text.JsConfig.EmitLowercaseUnderscoreNames


mythz:
This PR would change it to not reflect the true state of the DTO where Services would no longer be able to detect if an empty array or an absence of value was sent so it needs to be opt-in which I've added in [this commit](https://github.com/ServiceStack/ServiceStack/commit/c87decd22bb379d806beb0d05d3a274e348a541e).

Which you can skip for empty arrays with:

```csharp
ClientConfig.SkipEmptyArrays = true;
```

src/ServiceStack.Client/UrlExtensions.cs
```diff
@@ -481,7 +481,7 @@ internal static string GetQueryString(object request, IDictionary<string, RouteM
                     continue;
 
                 var value = queryProperty.Value.GetValue(request, true);
-                if (value == null)
+                if (value == null || (value is Array && value.ToJsv() == EmptyArray))
```
mythz: Note serializing an object to detect its serialized state is inefficient, can use `value is Array array && array.Length == 0` instead.

PR #1165, opened 2017-09-12
Typescript type generator does not properly output a default value for the C# bool type.

silkyfray:
Given the following route:
image

generating native types for typescript produces the following output:
image

the `createResponse ` return value should be `false`, not `new boolean()`


mythz:
Cheers 👍

PR #1170, opened 2017-10-27
Restore IsAccountLocked to protected from internal

Jothay:
This was previously protected (was using 4.0.40, updated to 4.5.14 today) and we had overridden this function and now can't because it's marked internal.


mythz:
Ok cool, it also needs to be accessible from outside the class so I've made it public. This change is now available from v5 which is now [available on MyGet](http://docs.servicestack.net/myget), please see [v5 changes](https://forums.servicestack.net/t/myget-pre-release-packages-upgraded-to-v5/4749) before upgrading.

PR #1173, opened 2017-11-16
Add Cloudflare IP

sundarchi1:
ServiceStack needs a way to get the real IP for Cloudflare visitors.
This may not be the right place to add it.. there's other classes where it may be appropriate. Feel free to close this request and submit your own if this is wrong.

### DIFF
--- a/src/ServiceStack.Interfaces/Web/IRequest.cs
+++ b/src/ServiceStack.Interfaces/Web/IRequest.cs
@@ -128,6 +128,11 @@ namespace ServiceStack.Web
         /// The Remote Ip as reported by X-Forwarded-For, X-Real-IP or Request.UserHostAddress
         /// </summary>
         string RemoteIp { get; }
+        
+        /// <summary>
+        /// The Remote Ip as reported by CF-Connecting-IP
+        /// </summary>
+        string CloudflareIp { get { return Headers["CF-Connecting-IP"]; } }
 
         /// <summary>
         /// The value of the Authorization Header used to send the Api Key, null if not available


sundarchi1:
Ok now I am very concerned. I clicked "Details" on that failed thing and it says "Attackers might be trying to steal your information" ????

Has submitting this PR put my computer at risk?!

mythz:
Static properties for CDN vendors is not something that belongs in a web services framework. If you want a typed property add a static extension method on `IRequest` or use a static class with const strings of the header names.

BoatLandIT:
Dear @mythz,

This is the IT team of the company @sundarchi1 works for. He filed a helpdesk ticket about a malicious website that may have compromised his working project. Out of abundance of caution, we wiped the potentially infected machine.

Our investigation revealed the following website was visited in that timeframe:
`https://build.servicestack.net/`

After visiting that site in a sandbox, we determined it is a simple case of an expired SSL certificate. I'm sure you can understand why some may react with concern if the warning says 'Attackers might be trying to steal your information".

While we understand the circumstances this pull request was closed, we would like to request that you fix the SSL certificate so our developers may use the website, without fear, to rectify any GitHub warnings that may show up in future pull requests.

..that is assuming that is an official ServiceStack website and not a phishing site. Please confirm this too. Additionally, please clarify what login credentials our developers would use to log in to the website.

mythz:
@BoatLandIT Right it's just an expired certificate as mentioned in Chrome's built-in warning. 

I've changed the server url back to http://build.servicestack.net so future PR's should report a `http` link, which means traffic is no longer encrypted but at least you wont see the warning dialog of the expired certificate. Please note that this is our internal CI server for our own CI purposes, i.e. it's not for external public access.

PR #1174, opened 2017-12-10
Revamped release notes

sundarchi1:
Lots of updates and corrections needed here.
Take care to remember to update this next time!


mythz:
The document is no longer valid, the release notes are published at: http://docs.servicestack.net/

PR #1176, opened 2017-12-14
GatewayExecuteAsync: don't apply response filters when applyFilters is false

migajek:
this fixes a bug in ServiceController.ExecAsync - when instructed not to apply filters, Respone filters are applied anyway (but not Request filters, making it even worse in my case).

I've discovered it when using Service Gateway (Sync and Async), that's why the tests are using Gateway and not comparing Exec/ExecAsync directly. I suppose they could be made shorter but I believe they still show the point.

without the fix, the second test (using Gateway.SendAsync) fails as the list contains 3 elements (ResponseFilter is called for Internal Request made through Gateway)


mythz:
great, thx for the fix + tests!

PR #1171, opened 2017-11-05
Added support for async PushProperty for available loggers 

snakefoot:
Log4net + NLog + Serilog

Help resolve https://forums.servicestack.net/t/how-to-provide-logger-with-custom-parameters-attributes/4828

PR #1178, opened 2018-01-16
Fixes NLog bug introduced when adding support for async PushProperty

snakefoot:
Fixes bug introduced with #1171 

See also https://forums.servicestack.net/t/wierd-error-trying-to-instantiate-nlogger-with-servicestack/5115


mythz:
Cheers 👍 this change is now available from v5.0.3 that's now [available on MyGet](http://docs.servicestack.net/myget).

PR #1179, opened 2018-01-17
Minor refactoring of the NLog logger unit test

snakefoot:
Better to listen for actual exceptions, instead of random warnings.


mythz:
Cheers 👍

PR #1181, opened 2018-02-08
Add unit test to show issue with 'HasKeys' on netcore.

Layoric:
This is an interesting case. On netcore, the `Request.QueryString.HasKeys()` in a service will return false even when keys are present and accessible from `Request.QueryString.AllKeys`.

This works fine if manually creating a `NetCoreQueryStringCollection` and using `HasKeys` but via the request pipeline, it doesn't.

This PR only adds the test to show it's a problem.

### DIFF
--- a/tests/NetCoreTests/ContainerTests.cs
+++ b/tests/NetCoreTests/ContainerTests.cs
@@ -1,8 +1,8 @@
-﻿using System;
+﻿using System.Net.Http;
+using System.Threading.Tasks;
 using Funq;
 using NUnit.Framework;
 using ServiceStack;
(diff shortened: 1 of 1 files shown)


Layoric:
Forgot to mention, this only became an issue when migrating from 1.0.44 Core packages.

Looks like this change to a decorator might from an interface implementation might be the cause.

https://github.com/ServiceStack/ServiceStack/commit/19ed926308ae61fced02b464d535677e09ab6a59#diff-47ceae23a05c053dd41ee339ef995ecc

The `public new bool HasKeys()` is never getting called.

PR #1182, opened 2018-02-18
Added ReturnAttribute for use in OrmLite

madaleno:
Added the ReturnAttribute to implement in OrmLite.
Properties marked with this attribute will be included in the RETURNING (FireBird and PostgreSql) or OUTPUT (SqlServer) clause when building sql sentences for INSERT or UPDATE operations.

This allows to use SEQUENCES in SqlServer for Id columns instead of IDENTITY.
Also and not less useful, allows to get back field values that might be updated in database triggers when doing the same operations.

### DIFF
--- a/src/ServiceStack.Interfaces/DataAnnotations/ReturnAttribute.cs
+++ b/src/ServiceStack.Interfaces/DataAnnotations/ReturnAttribute.cs
@@ -1,0 +1,26 @@
+
+using System;
+
+namespace ServiceStack.DataAnnotations
+{
+
+    /// <summary>
+    /// ReturnAttribute
+    /// Use to indicate that a property should be included in the 
+    /// returning/output clause of INSERT and UPDATE sql sentences
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
+    public class ReturnAttribute : AttributeBase {}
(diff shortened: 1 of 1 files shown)


mythz:
Are these meant to be property attributes? Do you have an example on how they may be used?

madaleno:
Yes. Just like the IgnoreAttribute.

Implementing it in OrmLite atm but the UpdateAttribute might be hard to implement if possible.
Already implemented the InsertAttribute for FireBird and SqlServer dialects.

Example:
class Person {
[Return]
[AutoIncrement]
[PrimaryKey]
public int Id { get; set; }
public string Name { get; set; }
[Return]
[Compute]
// field value calculated by a the database trigger.
public int SomeField { get; set; }
};

this would generate this sql statement in FireBird:
`insert into person (name) values (:name) returning id, somefield;`
and this one in sqlserver:
`insert into person (name) output inserted.id, inserted.somefield values (:name);`

`var p = new Person { Name = "me" };
d.Insert(p);`
// p.Id and p.SomeField should now have the new inserted value at db level.

mythz:
ok thx, i'll wait for the OrmLite PR to see what it ends up looking like.

PR #1183, opened 2018-02-22
Removed unused Return attributes

madaleno:
Removed unused Return attributes.


mythz:
Can you leave `[Return]` and remove the others, there's no need to be unnecessarily verbose.

madaleno:
ok.

madaleno:
Done. 
But I think "Return" is a little ambiguous and doesn't clarify the real use case for this attribute.

PR #1185, opened 2018-03-14
Support inlining of schemas in OpenApi spec

georgehemmings:
Added `InlineSchema` property to `ApiAttribute`.

When true, the schema will be defined inline with the parameter rather than in the definitions then referenced.

Discussed in forum: https://forums.servicestack.net/t/inline-declarations/4788

### DIFF
--- a/src/ServiceStack.Api.OpenApi/OpenApiFeature.cs
+++ b/src/ServiceStack.Api.OpenApi/OpenApiFeature.cs
@@ -39,12 +39,15 @@ namespace ServiceStack.Api.OpenApi
 
         public List<string> AnyRouteVerbs { get; set; }
 
+        public List<string> InlineSchemaTypesInNamespaces { get; set; }
+
         public bool DisableSwaggerUI { get; set; }
 
         public OpenApiFeature()
         {
             Tags = new List<OpenApiTag>();
             AnyRouteVerbs = new List<string> { HttpMethods.Get, HttpMethods.Post, HttpMethods.Put, HttpMethods.Delete };
+            InlineSchemaTypesInNamespaces = new List<string>();
         }
 
         public void Configure(IAppHost appHost)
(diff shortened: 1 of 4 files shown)


mythz:
I'd prefer it was a global setting on `OpenApiFeature` rather than having to specify it on every individual `[Api]` Attribute. Trying to support adhoc uses on both will cause more issues and requires more effort for end users to use, I'm also hoping the PR impl will be simplified with a global setting.

Why is `ClrType` needed? I don't like how disruptive it is and how ResultScope needs to be customized.

georgehemmings:
A global setting doesn't provide enough control on it's own, personally I'd want all the request and response DTOs in-lined (as they are single use for one operation), but also **some** "types" that are only used for input. There could be global settings to define the defaults for requests, responses and types, but I'd still need a mechanism to allow type-by-type control of in-lining. Instead of a property on the attribute, it could be a filter/fn on the `OpenApiFeature`, if you'd prefer? 

`ClrType` is used to relate the `OpenApiSchema` to the type it was generated from, as the `definitions` [should only contains referenced types](https://github.com/georgehemmings/ServiceStack/blob/86fd0f7c3b7dddb6d0d40ffd35ffabee3fc7c9ee/src/ServiceStack.Api.OpenApi/OpenApiService.cs#L85) (not in-lined). This could be a map instead of a property I suppose.

`ResultScope` needs to be customised because `Items` on `OpenApiDataTypeSchema` [is declared as a  object dictionary](https://github.com/georgehemmings/ServiceStack/blob/86fd0f7c3b7dddb6d0d40ffd35ffabee3fc7c9ee/src/ServiceStack.Api.OpenApi/Specification/OpenApiDataTypeSchema.cs#L18). This results in `__type` field getting added to the JSON, which isn't valid as the schema is defined by the OpenApi specification. I don't think preventing the addition of type info is a concern, as the aim is to produce JSON that is compatible with the spec.

georgehemmings:
Having thought about this more, I don't actually need type-by-type control after all, namespace control is enough for me.

``` csharp
Plugins.Add(new OpenApiFeature
{
    InlineSchemaTypesInNamespaces = new List<string>
    {
        "Project.ServiceModel",
    }
});
```

I can still add a global `InlineSchema` if you want, but namespace is pretty close anyway. Please let me know.

mythz:
ok thanks, FYI this PR is now available from v5.0.3 that's now [available on MyGet](http://docs.servicestack.net/myget).

PR #1186, opened 2018-04-01
NLog with NetStandard support

snakefoot:
Upgraded to NLog ver. 4.5


mythz:
Cheers, FYI if you need it, the .NET Standard 2.0 build of NLog is now available from v5.0.3 that's [now on MyGet](http://docs.servicestack.net/myget).

PR #1187, opened 2018-05-05
RequestCompressionType is not considered when using AsyncServiceClient

tomfroehle:
**Problem:** When setting the RequestCompressionType on a JsonServiceClient and using the SendAsync-method a SerializationException occurs because the Content-Encoding header is not set in the WebRequest

**Proposed solution in this PR:**
I added the Property RequestCompressionType to the AsyncServiceClient and set it when setting the Property in the ServiceClientBase. The RequestCompressionType is then (anologous to the ServiceClientBase) added as a header before sending the WebRequest. I added tests to the AsyncServiceClientTests and the ZipServiceClientTests

### DIFF
--- a/src/ServiceStack.Client/AsyncServiceClient.cs
+++ b/src/ServiceStack.Client/AsyncServiceClient.cs
@@ -91,6 +91,8 @@ namespace ServiceStack
         public string BaseUri { get; set; }
         public bool DisableAutoCompression { get; set; }
 
+        public string RequestCompressionType { get; set; }
+
         public string UserName { get; set; }
 
         public string Password { get; set; }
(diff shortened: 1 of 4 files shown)


tomfroehle:
I noticed that in the AsyncServiceClientTests there are Fixtures for the different Client-Types (e.g. XmlAsyncServiceClientTests, JsonAsyncServiceClientTests). They all override the abstract CreateServiceClient()-method but this method is never used. The tests all instantiate a JsonServiceClient. Wasn't sure whats your policy on finding stuff like this while fixing another issue.

mythz:
Thanks for the PR 👍 I've continued a [response on the forums](https://forums.servicestack.net/t/compression-in-netstandard/5694/2?u=mythz).

tomfroehle:
Thank you for the quick feedback!

PR #1188, opened 2018-06-25
Make prefetch configurable

DeonHeyns:
This pull request allows consumers to set the prefetch count but still defaults to the most optimal as described in:
http://www.rabbitmq.com/blog/2012/04/25/rabbitmq-performance-measurements-part-2/
http://www.rabbitmq.com/amqp-0-9-1-reference.html


mythz:
Looks good to me, thanks Deon 👍

PR #1191, opened 2018-08-04
Marked addIndex field as volatile

dmytroliubarskyi:
Marked _addIndex_ field (accessed by multiple threads concurrently) as **volatile** to suppress just-in-time optimization around it performed by next generation of JIT (RyuJIT) and causing concurrency issues (endless loop in Take() method under high-load).


neoeinstein:
Knowing that v3 is officially unsupported, would it be possible to release a 3.9.72 version with this concurrency fix?

PR #1192, opened 2018-08-17
Fix copy-paste bug in ProxyFeatureHandler

rfortunatov-softheme:
From my understanding this is a copy-paste issue and TransformRequest should be checked before execution.


mythz:
Nice catch, thx! This change is running through CI, will let you know when it's available on MyGet.

mythz:
This change is available from v5.1.1 that's now [available on MyGet](http://docs.servicestack.net/myget).

rfortunatov-softheme:
Great, thanks.

PR #1194, opened 2018-09-27
Add a flag to relax treating warnings/info as errors

DeonHeyns:
With FluentValidation you can set the Severity level to Info, Warning or Error using the WithSeverity extensions method.

The ValidationFeature does not distinguish between the different severity levels and will return an error response back to the caller regardless of the severity set.

This PR adds a flag to disable this behavior. 

Instead of getting an error response returned, if the response DTO inherits from IHasResponseStatus, the response status property will be populated with the failed validations and the meta dictionary property will contain the severity.

### DIFF
--- a/src/ServiceStack/Validation/ValidationFilters.cs
+++ b/src/ServiceStack/Validation/ValidationFilters.cs
@@ -104,35 +104,28 @@ namespace ServiceStack.Validation
         {
             var ruleSet = req.Verb;
 
-            try
-            {
-                ValidationResult validationResult;
+            ValidationResult validationResult;
 
-                if (validator.HasAsyncValidators(ruleSet))
-                {
-                    validationResult = await validator.ValidateAsync(
-                        new ValidationContext(requestDto, null, new MultiRuleSetValidatorSelector(ruleSet))
-                        {
-                            Request = req
-                        });
-                }
-                else
-                {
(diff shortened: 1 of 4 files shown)


src/ServiceStack/Validation/ValidationFilters.cs
```diff
@@ -62,9 +65,74 @@ public static async Task RequestFilterAsync(IRequest req, IResponse res, object
...
+            }
+
+            response.ResponseStatus = responseStatus;
+        }
+
+        private static async Task<ValidationResult> Validate(IValidator validator, IRequest req, object requestDto)
+        {
+            var ruleSet = req.Verb;
+
+            try
```
mythz: You don't need `try { ... } finally { using(validator as IDisposable){} }` You can just use `using (validator as IDisposable) { ... }`

mythz:
Can you also include a test that tests the new behavior.

src/ServiceStack/Validation/ValidationFilters.cs
DeonHeyns: Removed the try finally block

DeonHeyns:
Added tests to show new and old behavior still working

mythz:
Cheers 👍

PR #1196, opened 2018-10-16
Auto Batched Requests - share current req index with rest of pipeline

georgehemmings:
When a MultiRequest is being processed it would be useful for the rest of the pipeline to be able to obtain the index of the request within the batch. For example if an error occurs a `Meta` item could be added to the `ResponseStatus` which indicated the position that the error relates to.

This PR adds `MultiRequestIndex` as a req item that can be retrieved anywhere in the pipeline. If this is acceptable I'll add tests and refine the implementation.


mythz:
Please rename it to AutoBatchIndex so it’s tied to the feature.

mythz:
Cheers!

PR #1197, opened 2018-10-18
Auto Batched Requests - share current req index with rest of pipeline part 2

georgehemmings:
Further to #1196 

* Add AutoBatchIndex for global response filters
* Add tests


mythz:
Cheers!

PR #1198, opened 2018-10-19
Add AutoBatchIndex to ResponseStatus.Meta for auto batched requests

georgehemmings:
This allows the client to relate the error to the request which caused the error.

### DIFF
--- a/src/ServiceStack/Host/ServiceRunner.cs
+++ b/src/ServiceStack/Host/ServiceRunner.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


georgehemmings:
Yes. However in some circumstances X-AutoBatch-Completed does not relate to the index of the request object that caused the error. For example during validation the X-AutoBatch-Completed will always be null.

Putting AutoBatchIndex in Meta provides a consistent place to check to see which request object the error relates to.

mythz:
Ok no worries, thx.

georgehemmings:
@mythz up to now I've been checking out the 5.4 tag as a starting pointing, as that builds. If I checkout master what do I need setup to get it to build correctly?

mythz:
Yeah basically you need to checkout the other major repos, build a Release version then run lib\copy.bat to copy over the latest builds in each repo. 

On every version we push the latest .dlls in each repo which is how each repo is able to build stand-alone.

PR #1199, opened 2018-10-22
Add ext method for use in Custom Batched Requests Implementations

georgehemmings:
Using this method to enumerate the requests batch ensures the same behaviour
as regular (non-custom) batch requests. Namely, the AutoBatchIndex item is
set correctly.

The new test shows usage with a custom batched request implementation. If the custom impl performs some kind of complex validation then performing this within the EachRequest "loop" will ensure the exceptions ResponseStatus.Meta is set as with non-customs.

### DIFF
--- a/src/ServiceStack/HttpRequestExtensions.cs
+++ b/src/ServiceStack/HttpRequestExtensions.cs
@@ -1111,5 +1111,19 @@ namespace ServiceStack
                     throw new NotSupportedException($"Unknown IHttpRequest property '{name}'");
             }
         }
+
+        public static void EachRequest<T>(this IRequest httpReq, Action<T> action)
+        {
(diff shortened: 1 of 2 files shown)


src/ServiceStack/HttpRequestExtensions.cs
```diff
@@ -1111,5 +1111,19 @@ public static string GetRequestValue(this IHttpRequest req, string name)
...
+        {
+            if (!(httpReq.Dto is IEnumerable<T> requests))
+                return;
+
+            requests.Each((i, dto) =>
+            {
+                httpReq.Items["AutoBatchIndex"] = i;
+                action(dto);
+            });
+
```
mythz: Ideally you should avoid LINQ in library code for reduced allocations, e.g. using `foreach` or `for` .

PR #1200, opened 2018-11-01
Unexpected end of Stream when uploading to aspnet core

AnisAli:
At the time of RequestLogFeature Register, SS is attaching the Callback method for reading InputStream Buffer at HTTP listener level for all content type. That callback method creating the issue when the content type if multipart/form-data at least in .netcore 2.0

https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack/Host/HttpListener/ListenerRequest.cs  *line 221*

More Details about issue
https://forums.servicestack.net/t/unexpected-end-of-stream-when-uploading-to-aspnet-core/6478/2

Even the following commit will not solve the issue, because of InputStream already read before this GetRawBody call in InMemoryRollingRequestLogger.cs
https://github.com/ServiceStack/ServiceStack/commit/166e785b2dfce8a29ccfc30d2e172c69728cf6c3

### DIFF
--- a/src/ServiceStack/RequestLogsFeature.cs
+++ b/src/ServiceStack/RequestLogsFeature.cs
@@ -118,10 +118,27 @@ namespace ServiceStack
 
             if (EnableRequestBodyTracking)
             {
+#if NETSTANDARD2_0
+                appHost.PreRequestFilters.Insert(0, (httpReq, httpRes) =>
+                {
+                    if (httpReq.ContentType != null)
+                    {
+                        if (!(httpReq.ContentType.MatchesContentType(MimeTypes.MultiPartFormData)))
+                        {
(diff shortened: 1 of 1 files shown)


mythz:
I've gone with the simpler impl at https://github.com/ServiceStack/ServiceStack/commit/66fc27b3bd35032fad28038bf40e93bcd5de9815

AnisAli:
> I've gone with the simpler impl at [66fc27b](https://github.com/ServiceStack/ServiceStack/commit/66fc27b3bd35032fad28038bf40e93bcd5de9815)

@mythz 
this line will throw exception in case of null ContentType- We already face this issue. So please makesure it properly handle the null content type. 
httpReq.ContentType.MatchesContentType(MimeTypes.MultiPartFormData))

PR #1202, opened 2019-01-21
Add BadRequest HttpError factory method

georgehemmings:


mythz:
Cheers 👍 

This is now available from the latest v5.4.1 that's now on MyGet.

PR #1204, opened 2019-03-11
Fix memory leak: After disposing ServiceStackHost is kept in memory

todorm85:
Even after ServiceStackHost has been disposed, the static event TaskScheduler.UnobservedTaskException holds a reference to the object and thus it is kept in memory throughout the lifetime of the application. This creates a problem with systems that have "soft" restarts where the app domain is never disposed, and ServiceStack is reinitialized multiple times.

### DIFF
--- a/src/ServiceStack/ServiceStackHost.cs
+++ b/src/ServiceStack/ServiceStackHost.cs
@@ -775,19 +775,21 @@ namespace ServiceStack
 
             if (config.LogUnobservedTaskExceptions)
             {
-                TaskScheduler.UnobservedTaskException += (sender, args) =>
-                {
-                    args.SetObserved();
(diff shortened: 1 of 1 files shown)


mythz:
Great, thx for the PR. Can you [approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so we can merge this, thx.

todorm85:
Sorry about that, I thought I`ve signed it already. It should be ok now.

mythz:
Cheers 👍 This change is now available from the [latest v5.4.1 now available on MyGet](https://docs.servicestack.net/myget).

PR #1205, opened 2019-03-19
Request Logging: Ensure that an ErrorResponse is logged as an error.

EricMunn:
Responses of type ErrorResponse are not logged to the error log.

1. **HttpExtensions.EndHttpHandlerRequest** and **EndHttpHandlerRequestAsync** do not include the response in the call to **IRequestLogger.Log**, thus downstream handlers see the response as null (this also appears to prevent non-error responses from being logged when EnableResponseTracking is true),
2. **HttpResultUtils.IsErrorResponse** only considers **IHttpError** or **Exception** to be an error response, and thus
3. **InMemoryRollingRequestLogger.ToSerializableErrorResponse** does not serialize an ErrorResponse as an error.

Added tests fail in pre-PR code.

First PR, so comments/changes/criticisms welcome.

### DIFF
--- a/src/ServiceStack/Host/InMemoryRollingRequestLogger.cs
+++ b/src/ServiceStack/Host/InMemoryRollingRequestLogger.cs
@@ -191,6 +191,8 @@ namespace ServiceStack.Host
         {
             if (response is IHttpResult errorResult)
                 return errorResult.Response;
+            else if (response is ErrorResponse errorResponse)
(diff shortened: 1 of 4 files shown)


mythz:
Awesome work, thx for the PR + tests! 👍

mythz:
FYI this change is now available from v5.4.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1206, opened 2019-04-25
added netstandard2.0 support

VagyokC4:
Was there a reason why you don't support netstandard2.0 with this project?


mythz:
Thanks, I've also upgraded the deps in this commit: https://github.com/ServiceStack/ServiceStack/commit/9bda4ec3176f309c086e3f80bcb70c72e0b1b3c1

The `netstandard2.0` builds for `ServiceStack.Authentication.MongoDb` are available from v5.5.1 that's [now on MyGet](https://docs.servicestack.net/myget).

PR #1210, opened 2019-11-13
NLogLogger - Allow NLog to capture Exception-object when only parameter

snakefoot:
Give Exception-object to NLog when called like this:

`Log.Fatal(ex);`


mythz:
Cheers 👍

PR #1211, opened 2019-11-13
NLogLogger - Allow NLog to capture Exception-object when only parameter (unit test)

snakefoot:
Unit test for #1210 (Requires NLog 4.5.4 that includes https://github.com/NLog/NLog/pull/2586)


mythz:
Cheers 👍

snakefoot:
Fast as always. Thank you for the swift merges.

mythz:
np, FYI this change is now available from the latest v5.7.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

juniglee:
Hi @mythz , is this available on NuGet yet?

mythz:
@juniglee No only major releases get published to NuGet (which wont be available in the immediate future), you'll need to use v5.7.1 packages on MyGet or copy the source files and use your local version.

PR #1213, opened 2019-12-07
Neo4j IAuthRepository & Neo4j ICacheClient

neildobson-au:
Impl + tests complete. Just need some doco.

Was thinking of adding a readme in the same directory and use one of the sample projects such as vue-spa to walkthrough how to swap out the existing auto repo with this one, how to install Neo4j etc.


neildobson-au:
I've added Neo4jCacheClient which implements the full & extended caching interfaces. Cache tests are passing except those which call `Cache.Get<IAuthSession>`. Internally cache data is serialized to JSV and returned using `JsvStringSerializer().DeserializeFromString<T>(cacheEntry.Data)` however I'm not sure how one can resolve a type via its interface without storing the concrete type along with the data. Some tests however are doing exactly this. Any ideas?

neildobson-au:
FYI all (de)serialization duties are using ServiceStack.Caching.Neo4j.CacheValueExtensions which uses ServiceStack.Text.

neildobson-au:
I've just updated the [Neo4j.Driver](https://www.nuget.org/packages/Neo4j.Driver.Simple.Signed/) packages to support Neo4j 4.0. Unfortunately this resulted in breaking changes not only in their APIs but also they have dropped support for NET Framework (at least < 4.6.1).

This means integrating is harder due to some of the ServiceStack projects (particularly the test ones) still supporting multiple targets including 4.5 / 4.6 etc.

For this reason I'll abandon the PR for now. The work is done and I'll maintain a separate set of nuget packages for auth / cache providers.

Once things have settled down a bit I'll attempt another PR in future.

Thanks.

PR #1214, opened 2020-01-21
Refresh BearerToken properly in JsonHttpClient

joncinque:
JsonHttpClient sends a refresh request if `RefreshToken` is set, but
while processing the refresh, the underlying `HttpClient` is not updated,
so subsequent requests with the same JsonHttpClient will again
need to refresh the out-of-date `BearerToken`.  The update to
`BearerToken` is never reflected in `HttpClient`.  Note that this only
happens with token-only authorization.  Cookie and Basic auth properly
update the `HttpClient`.

There are also some whitespace deletions in this PR, but those can be reverted.
Let me know if you have any questions.

### DIFF
--- a/src/ServiceStack.HttpClient/JsonHttpClient.cs
+++ b/src/ServiceStack.HttpClient/JsonHttpClient.cs
@@ -58,7 +58,7 @@ namespace ServiceStack
         public string BearerToken { get; set; }
         public string RefreshToken { get; set; }
         public string RefreshTokenUri { get; set; }
-        
+
         public bool UseTokenCookie { get; set; }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/ServiceStack.HttpClient/JsonHttpClient.cs
```diff
@@ -299,12 +299,13 @@ public Task<TResponse> SendAsync<TResponse>(string httpMethod, string absoluteUr
                                         else
                                         {
                                             refreshRequest.AddBearerToken(this.BearerToken = accessToken);
+                                            client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", accessToken);
```
joncinque: This is the substantive line of this PR

mythz:
thx for the PR!

If needed this change is available from the latest v5.8.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

joncinque:
Great, thanks for the quick response!

PR #1219, opened 2020-02-22
Allow for local build to default to the latest build available on build machine

Siliconrob:
The current batch build is constructed off the assumption of using Visual Studio 2017.  This change allows for the system to find the latest MSBuild.exe on the machine and then build ServiceStack locally from that.

Ran into this issue while trying to build a local copy of ServiceStack dlls ( I want to use them in open source development rather than nuget libraries to avoid license issues and ran into this issue)


Siliconrob:
Had to update to latest `5.8.1` for build target as kept running into dead link on ServiceStack.OrmLite access to feed 

This one in particular `<add key="ServiceStack CI feed" value="https://build.servicestack.net/guestAuth/app/nuget/v1/FeedService.svc" />`

mythz:
ok cheers thx for the MSBuild detection script!

PR #1220, opened 2020-03-06
Add IOC GetServiceEntryNamed<T> and ExistsNamed<T>

jezzsantos:
As an extension to the previous commit, just adding these overloads for GetServiceEntryNamed<T> and ExistsNamed<T> that are needed for named registrations.


mythz:
Cheers 👍

PR #1221, opened 2020-04-07
Fix broken grpc tests

mgravell:
1. fix AddField typo
2. use AfterApplyDefaultBehaviour to move timing of type configuration (moves config code to GrpcConfig)
3. above needs newer protobuf-net build

Main change is 1; this seems to have been a change when `AddField` was refactored after my earlier PR; [original code](https://github.com/ServiceStack/ServiceStack/commit/14dd3eb9d29661c7990c7f30c8508d9bfc1a988d#diff-eeb5a8117e8744e028e08c16d2c43d97R746) added `field.ItemType`; code in "master" added `field.ItemType ?? field.MemberType`, which incorrectly tries to configure everything as "repeated". If this was done for a specific reason, let me know - I can look into that.

2. moves type config to use the newer protobuf-net API that allows configuration *during discovery*, avoiding a lot of timing problems; this required moving code from `GrpcServiceClient` to `GrpcConfig`

There's still something going on re `CheckServiceProto_DerivedType` - will look more this later today.

### DIFF
--- a/src/ServiceStack.Extensions/GrpcFeature.cs
+++ b/src/ServiceStack.Extensions/GrpcFeature.cs
@@ -310,15 +310,7 @@ namespace ServiceStack
             // All DTO Types with inheritance need to be registered in GrpcMarshaller<T> / GrpcUtils.TypeModel
             foreach (var dto in allDtos)
             {
-                try
-                {
-                    GrpcConfig.Register(dto);
-                }
-                catch (Exception e)
-                {
-                    Console.WriteLine(e);
-                    // throw;
-                }
+                GrpcConfig.Register(dto);
             }
         }
 
--- a/src/ServiceStack.Extensions/ServiceStack.Extensions.csproj
+++ b/src/ServiceStack.Extensions/ServiceStack.Extensions.csproj
@@ -12,7 +12,7 @@
     <ItemGroup>
         <PackageReference Include="Grpc.AspNetCore.Server" Version="2.28.0-pre2" />
         <PackageReference Include="Grpc.Core.Api" Version="2.28.0-pre3" />
-        <PackageReference Include="protobuf-net" Version="3.0.0-alpha.133" />
+        <PackageReference Include="protobuf-net" Version="3.0.0-alpha.143" />
         <PackageReference Include="protobuf-net.Grpc.AspNetCore" Version="1.0.36" />
         
         <PackageReference Include="System.Threading.Channels" Version="4.7.0" />
--- a/src/ServiceStack.GrpcClient/GrpcConfig.cs
+++ b/src/ServiceStack.GrpcClient/GrpcConfig.cs
@@ -2,12 +2,87 @@ using System;
 using System.Collections.Concurrent;
 using System.Reflection;
 using ProtoBuf.Meta;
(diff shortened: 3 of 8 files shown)


mythz:
thx Marc, the refactor was only me trying out different things to try resolve issues. I'll check this out now, thx.

mgravell:
The key point on 2 is that the flow is:

- check for existing meta-type
- if there, check for lock and return it
- if not there:
  - create locked
  - (new event: before configure)
  - perform auto-discovery if required
  - (new event: after configure)
  - release lock, return type

this means that you can now hook type customization *while it is locked*, which prevents other threads getting hold of the meta-type and fighting over it (or worse: creating a serializer, which freezes the type for further changes)

mythz:
Cool so is the locking happening behind the scenes when `AfterApplyDefaultBehaviour` is invoked?

mgravell:
the locking happens when doing anything that fetches an *existing* meta-type, i.e. `someModel[type]` etc; essentially this includes:

``` c#
if (metaType.Pending) WaitOnLock();
```

which synchronizes against the model, so only one thread can be mutating the model at a time; the lock here is re-entrant, so you can't deadlock yourself

mythz:
Can confirm this resolved most failing tests just `CheckServiceProto_DerivedType` is failing when the test suite is run, passes when run individually.

I've just enabled more stress testing with the porting of more tests of a [new AutoCrud feature](https://forums.servicestack.net/t/autocrud-preview/8298/41) to gRPC and happy to report all tests pass except for `Can_CreateRockstarAudit` which uses shadowed properties and has peculiar behavior so I thought I'd make you aware of it. 

It's not something that needs fixing as can say shadowed properties isn't supported but you might be interested in its behavior, where if you:

  1. Comment out the `LivingStatus` property the test passes 
  2. When using 1 & 2 index for `Id`, `FirstName` properties it throws `RpcException: Status(StatusCode=Unknown, Detail="Exception was thrown by handler.")`
  3. But when Index changed to 11,12 it causes empty DTO to be sent (i.e. all properties empty)

```csharp
[DataContract]
public abstract class RockstarBase
{
    [DataMember(Order = 1)]
    public string FirstName { get; set; }
    //...
    [DataMember(Order = 6)]
    public LivingStatus LivingStatus { get; set; }
}
[DataContract]
public class UpdateRockstarAudit : RockstarBase, IPatchDb<RockstarAudit>, IReturn<EmptyResponse>
{
    [DataMember(Order = 1)]
    // [DataMember(Order = 11)]
    public int Id { get; set; }
    [DataMember(Order = 2)]
    // [DataMember(Order = 12)]
    public new string FirstName { get; set; }
    [DataMember(Order = 13)]
    public new LivingStatus? LivingStatus { get; set; }
}
```

I've tried creating a repro in [Can_serialize_hidden_property()](https://github.com/ServiceStack/ServiceStack/blob/ae3e55a9d8f0af7e431319a92f6c829b04764733/tests/ServiceStack.Extensions.Tests/ProtobufTests.cs#L147) using the ProtoBuf Serializer directly and it passes, so might be an interop issue, anyway doesn't need fixing just wanted to let you know about the corner-case behavior.

PR #1225, opened 2020-06-02
#Script - Option to parse numbers as decimals

georgehemmings:
I'm considering using #Script/Js.eval as an embedded language in a financial application, for user defined formulas. However currently all "decimal" numbers are parsed as `double`s. I appreciate this is in keeping with the JavaScript roots.

It would be great if there was an opt-in option to parse numbers as `decimal`s instead of `double`s.

I've added a static flag to the `ScriptContext`, but this could be an instance field, it would just need passing all the way done 🐢.

Please let me know if this is something you'd consider allowing, if it is I can look at adding tests and changing to an instance field etc.

### DIFF
--- a/src/ServiceStack.Common/Script/ScriptConfig.cs
+++ b/src/ServiceStack.Common/Script/ScriptConfig.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using ServiceStack.IO;
+using ServiceStack.Text;
 
 namespace ServiceStack.Script
 {
@@ -36,7 +37,8 @@ namespace ServiceStack.Script
         public static string DefaultErrorClassName { get; set; } = "alert alert-danger";
         public static bool AllowUnixPipeSyntax { get; set; } = true;
         public static bool AllowAssignmentExpressions { get; set; } = true;
-        
+        public static ParseRealNumber ParseRealNumber = numLiteral => numLiteral.ParseDouble();
+
         public static CultureInfo CreateCulture()
         {
             var culture = DefaultCulture;
@@ -52,4 +54,6 @@ namespace ServiceStack.Script
             return culture;
         }
     }
+
+    public delegate object ParseRealNumber(ReadOnlySpan<char> numLiteral);
 }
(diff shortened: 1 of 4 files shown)


mythz:
Any static config should be on `ScriptConfig`. 

But I'm not a fan of the bespokeness, i.e. if you want to enable customization it should be open for generic functionality so you could instead override how all numbers are parsed:

```csharp
ScriptConfig.ParseNumber = numLiteral => numLiteral.ParseDecimal();
```

Or if you only want to override real numbers:

```csharp
ScriptConfig.ParseRealNumber = numLiteral => numLiteral.ParseDecimal();
```

mythz:
Please also include tests for any new feature.

mythz:
But would it be better if there was just a literal for decimal like in C# e.g. `1.0m`

As the custom static configuration means your scripts wont be portable across a different context using the default configuration.

georgehemmings:
I had totally missed `ScriptConfig` somehow 🤦 

For my scenario, simple mathematical expressions written by end-users, anything "weird" like "m" suffix after numbers isn't desirable.

I'll look into implementing it like this: `ScriptConfig.ParseRealNumber = numLiteral => numLiteral.ParseDecimal();`

mythz:
Yep np. Adding `ScriptConfig.ParseRealNumber` is going to be easiest wait to impl atm, 

Though if there's going to be a lot more cases where we'd want to be able to change AST behavior we'd want it to be configurable on the `ScriptContext` instance which would pass something like an `AstContext` into the parser/tokenizer fn's which can optionally be enabled via a directive (e.g. `#pragma decimal`), but that's going to require a lot more work.

src/ServiceStack.Common/Script/ScriptConfig.cs
```diff
@@ -36,7 +37,8 @@ public static class ScriptConfig
         public static string DefaultErrorClassName { get; set; } = "alert alert-danger";
         public static bool AllowUnixPipeSyntax { get; set; } = true;
         public static bool AllowAssignmentExpressions { get; set; } = true;
-        
+        public static ParseRealNumber ParseRealNumber = numLiteral => numLiteral.ParseDouble();
```
georgehemmings, line 40: Couldn't define as Func<ReadOnlySpan<char>, object> because "ReadOnlySpan may not be used as a type argument for generic delegates"

https://stackoverflow.com/questions/53155438/why-readonlyspan-may-not-be-used-as-a-type-argument-for-generic-delegates-and-ge

georgehemmings:
I guess you've been busy with the release :-), but would you accept this now?

mythz:
Apologies, missed the notification. Thought the the commit was still incoming.

PR #1226, opened 2020-06-18
Don't dispose HTTP message handler if it was injected

georgehemmings:
`JsonHttpClient` was disposing the injected `HttpMessageHandler` preventing re-use.

A `SocketsHttpHandler` instance pools connections so can be long-lived [without fear of socket exhaustion or DNS problems](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-3.1#alternatives-to-ihttpclientfactory).

For situations where the same `JsonHttpClient` can't be re-used (different authentication for example), it would be useful to at least re-use the handler and underlying connections (with cookies off).


mythz:
ok no worries, thx.

PR #1228, opened 2020-07-18
Change HttpMethod on SendAll and SendAllAsync based on IVerb interfaces

DeonHeyns:
With this pull request AutoBatched Requests is extended to take the HttpMethod marker of a request (IPost, IGet, IPatch, IDelete, IPut) in consideration when making the underlying request. 

Why: we use HttpVerbs for our messages to perform different actions. We'd like to use AutoBatched requests instead of having to call Put, Delete or Patch methods in a loop

### DIFF
--- a/src/ServiceStack.Client/AsyncServiceClient.cs
+++ b/src/ServiceStack.Client/AsyncServiceClient.cs
@@ -153,8 +153,7 @@ namespace ServiceStack
             this.PopulateRequestMetadata(request);
 
             var requestUri = absoluteUrl;
-            var hasQueryString = request != null && !HttpUtils.HasRequestBody(httpMethod);
-            if (hasQueryString)
+            if (!this.EmulateHttpViaPost && !HttpUtils.HasRequestBody(httpMethod) && request != null)
             {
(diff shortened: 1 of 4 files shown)


mythz:
Not sure if you have access to the failed tests in the link but this PR broke these tests: 
image

DeonHeyns:
Thanks @mythz I do not have access but will get these sorted

DeonHeyns:
I've changed how this works. Instead of forcing all callers into this new behavior rather let them opt in by setting EmulateHttpViaPost.

I could change this so that calling SendAll or SendAllAsync sets EmulateHttpViaPost to true and then resets it after or internally sets the XHttpMethodOverride based on calling SendAll. Thoughts @mythz

mythz:
ok cool, setting `EmulateHttpViaPost` explicitly as done in tests is preferred here.

DeonHeyns:
Cool - I’ll leave this PR open while I make my way through the other ServiceClients and update them when needed

DeonHeyns:
Reworked our APIs as this seemed a more semantic way to approach vs faking HTTP methods

PR #1229, opened 2020-08-19
fix for PATCH requests with multipart/form-data not populating the request DTO

dj-nitehawk:
a patch request with multipart/form-data such as the following results in the request dto objects not getting populated by servicestack.
```
curl --location --request PATCH 'http://localhost:8888/image' \
--form 'Width=1500' \
--form 'Height=1000' \
--form 'ID=5f3c03457118797a3a7a6f8c' \
--form 'File=@/D:/IMG_20190901_144155.jpg'
```

the issue seems to be a simple oversight/omission in  `HttpRequestExtensions.cs`

for more info please refer to my original SO question here: https://stackoverflow.com/questions/63473937

### DIFF
--- a/src/ServiceStack/HttpRequestExtensions.cs
+++ b/src/ServiceStack/HttpRequestExtensions.cs
@@ -552,7 +552,7 @@ namespace ServiceStack
                 map[name] = request.QueryString[name];
             }
 
-            if ((request.Verb == HttpMethods.Post || request.Verb == HttpMethods.Put)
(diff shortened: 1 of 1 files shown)


mythz:
Thx for the PR, although it's better to use the `HttpUtils.HasRequestBody()` API which I've updated it to use in this commit
https://github.com/ServiceStack/ServiceStack/commit/0fc6dd94a5cea2022cc66a4148ccc9dad1fb67b9

This change is now available from the latest v5.9.3 that's now [available on MyGet](https://docs.servicestack.net/myget)

PR #1230, opened 2020-09-20
Update AdminUsersFeature.cs

nukedbit:
I think you meant to check for permissions here :)


mythz:
Ahh yes, nice catch, thx 👍

PR #1231, opened 2020-12-16
Remove stack trace from SerializationException message

dwfrancis:
String interpolation for `SerializationException` calls `ex.ToString()` under the hood which inadvertently includes the stack trace in the error message.


mythz:
Cheers 👍

PR #1233, opened 2021-01-18
Make EventSubscription metadata immutable to prevent crashes when updated in the background

laurenceSaes:
The subscription metadata is updated during run-time which makes all existing enumerators on this object invalid. 

- The serializer iterates over all metadata fields. 
- This request fails when the metadata is updated in the background.

### DIFF
--- a/src/ServiceStack/ServerEventsFeature.cs
+++ b/src/ServiceStack/ServerEventsFeature.cs
@@ -542,6 +542,14 @@ namespace ServiceStack
 
         private readonly IResponse response;
         private long msgId;
+        
+        private Dictionary<string, string> _meta;
+              
+        public override Dictionary<string, string> Meta
+        {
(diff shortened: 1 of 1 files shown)


mythz:
Thx for the PR but have decided to change it to `ConcurrentDictionary` as the Meta gets updated in `OnCreated`/`OnConnect` callbacks to pass additional info to client SSE connection & we don't want to lose the updates.

This change is available from v5.10.5 that's now [available on MyGet](https://docs.servicestack.net/myget).

> For future PRs, can you please sign the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so they can be merged, thx.

PR #1238, opened 2021-01-30
Fix default LogFactory in NetCore

IkerCelorrio:
on NetCore with the UseServiceStack it delegates in Microsoft.Extensions.Logging ILogger if  LogManager.LogFactory == null, but LogManager.LogFactory could not be null because it coerce witha a new NullLogFactory(false)

https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack.Interfaces/Logging/LogManager.cs#L17-L21
```csharp
        public static ILogFactory LogFactory
        {
            get => logFactory ?? new NullLogFactory();
            set => logFactory = value;
        }
```


mythz:
thx for the PR 👍 FYI this change is now available from the latest v5.10.5 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1241, opened 2021-02-26
MS Graph: get id token and set roles in oauthprovider

DeonHeyns:


mythz:
Awesome, much appreciated, thx Deon 👍

DeonHeyns:
@mythz, you are too quick!

mythz:
This change is now available from the latest v5.10.5 that's now [available on MyGet](https://docs.servicestack.net/myget).

@DeonHeyns thanks again for the PR!

PR #1242, opened 2021-03-18
Introduce Asp.Net Core 3.1 & Net 5 targets

trejjam:
This PR adds new targets (`netcoreapp3.1;net5.0`) for `ServiceStack.Core` so we can drop following deprecated packages:
- `Microsoft.AspNetCore.Http@2.x`
- `Microsoft.AspNetCore.Http.Abstractions@2.x`
- `Microsoft.AspNetCore.Http.Extensions@2.x`
- `Microsoft.AspNetCore.Hosting@2.x`
- `Microsoft.AspNetCore.Hosting.Abstractions@2.x`
- `Microsoft.Extensions.Logging.Abstractions@2.x`
- `Microsoft.Extensions.DependencyInjection@2.x`

### DIFF
--- a/src/ServiceStack.Interfaces/Caching/ICacheClientAsync.cs
+++ b/src/ServiceStack.Interfaces/Caching/ICacheClientAsync.cs
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 namespace ServiceStack.Caching
 {
     public interface ICacheClientAsync
-#if NET472 || NETSTANDARD2_0        
+#if NET472 || NETSTANDARD2_0 || NETCOREAPP3_1 || NET5_0
         : IAsyncDisposable
 #endif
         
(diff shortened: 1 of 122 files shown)


mythz:
Yeah we definitely don't want to be polluting our code base with multiple custom builds for each concrete .NET runtime, which targeting .NET Standard 2.0 lets us avoid.

We'll reevaluate adding a new framework target when .NET 6 LTS is released. I'll leave this PR open incase anyone else wants any of these targets.

trejjam:
It is not primarily about targets but about getting rid of old `Microsoft.AspNetCore.*` packages which are from `2.x` era with the end of support at `2021-08-21`.

mythz:
Which we can look at doing after .NET 6 LTS, not interested in carrying runtime specific builds for every .NET Runtime version. Hopefully by then there'll be a new .NET Standard version we'll be able to target instead.

PR #1246, opened 2021-06-10
Use async cache client if registered to avoid possible NRE

Layoric:
Use async cache client if registered to avoid possible NRE when using `ICacheClientAsync`.


mythz:
nice catch, thx 👍

PR #1248, opened 2021-07-19
Added AccessToken to GetAccessTokenService

jezzsantos:
Return the access_token in the response even when `UseTokenCookie==true`

### DIFF
--- a/src/ServiceStack/Auth/JwtAuthProvider.cs
+++ b/src/ServiceStack/Auth/JwtAuthProvider.cs
@@ -493,7 +493,10 @@ namespace ServiceStack.Auth
             if (request.UseTokenCookie.GetValueOrDefault(jwtAuthProvider.UseTokenCookie) != true)
                 return response;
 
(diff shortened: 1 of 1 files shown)


mythz:
This would defeat the purpose of securing JWT's by using a HttpOnly Token Cookie. 

Set `UseTokenCookie=false` if you'd instead prefer for the token to be returned in the HTTP Request Body.

jezzsantos:
Hey @mythz, sorry I don't understand that design assumption.

Perhaps I'm forgetting something about security here.

HttpOnly cookies are to prevent XSS from reading cookie values. Which we definitely want, but I don't see what we are protecting against by not including the token in a HTTP response?

What do you see?

mythz:
Ajax refresh tokens requests use HttpOnly Token Cookies so their authenticated BearerToken cannot be read by JS which is why it's recommended they use HttpOnly Secure Cookies. Returning the JWT in the HTTP Body defeats this protection which would allow JS to access new authenticated tokens.

jezzsantos:
OK, thanks

PR #1250, opened 2021-09-10
Generic registration plugin with some tests.

Layoric:
This new plugin is to make it easier to have the functionality of `RegisterFeature` but with the ability to specify custom request DTO so data can be populated between custom `UserAuth` and custom `AuthUserSession`. Eg

```
class AppHost : AppSelfHostBase
{
    public AppHost()
        : base(nameof(GenericRegistrationFeatureTests), typeof(MyRegistrationService).Assembly) { }
    
    public override void Configure(Container container)
    {
        Plugins.Add(new AuthFeature(() => new CustomRegisterUserSession(),
            new IAuthProvider[]
            {
                new CredentialsAuthProvider()
            }));

        Plugins.Add(new GenericRegistrationFeature<MyRegistrationService,CustomRegister>());

        container.Register<IDbConnectionFactory>(new OrmLiteConnectionFactory(":memory:",
            SqliteDialect.Provider));
        container.Register<IAuthRepository>(new OrmLiteAuthRepository<CustomRegisterUserAuth,UserAuthDetails>(
            container.Resolve<IDbConnectionFactory>()));

        var authRepo = container.Resolve<IAuthRepository>();
        authRepo.InitSchema();
    }
}

public class MyRegistrationService : GenericRegisterService<CustomRegister>
{}

public class CustomRegister : Register, IReturn<RegisterResponse>
{
    public string NetworkName { get; set; }
}

public class CustomRegisterUserAuth : UserAuth
{
    public string NetworkName { get; set; }
}

public class CustomRegisterUserSession : AuthUserSession
{
    public string NetworkName { get; set; }
}
```

I'm not sure of the best way to reduce duplication between RegisterFeature and this feature so logic is shared without breaking existing `RegistrationFeature` usage. Same with the progression of this to a fully custom registration process, should PostAsync be overridable ? Is there maybe a better way to structure this so just the DTO is declared rather than the empty service, similar with AutoQuery and DTOs use of QueryDb<T>?

### DIFF
--- a/src/ServiceStack/Auth/GenericRegistrationService.cs
+++ b/src/ServiceStack/Auth/GenericRegistrationService.cs
@@ -1,0 +1,209 @@
+﻿using System;
+using System.Linq;
+using System.Globalization;
+using System.Threading.Tasks;
+using ServiceStack.FluentValidation;
+using ServiceStack.Text;
+using ServiceStack.Validation;
+using ServiceStack.Web;
+
+namespace ServiceStack.Auth
+{
+    [ErrorView(nameof(Register.ErrorView))]
+    [DefaultRequest(typeof(Register))]
+    public abstract class GenericRegisterService<T> : Service where T : Register
+    {
+        public static ValidateFn ValidateFn { get; set; }
+        
+        public static bool AllowUpdates { get; set; }
+
+        public IValidator<T> RegistrationValidator { get; set; }
+
+        public IAuthEvents AuthEvents { get; set; }
+        
+        /// <summary>
+        /// Create new Registration
+        /// </summary>
+        public async Task<object> PostAsync(T request)
+        {
+            var returnUrl = Request.GetReturnUrl();
+
+            var authFeature = GetPlugin<AuthFeature>();
+            if (authFeature != null)
+            {
+                if (authFeature.SaveUserNamesInLowerCase)
+                {
+                    if (request.UserName != null)
+                        request.UserName = request.UserName.ToLower();
+                    if (request.Email != null)
+                        request.Email = request.Email.ToLower();
+                }
+                if (!string.IsNullOrEmpty(returnUrl))
+                    authFeature.ValidateRedirectLinks(Request, returnUrl);
+            }
+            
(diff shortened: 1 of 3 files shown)


mythz:
Given the use-case we're trying to solve is to allow Users to provide their own Custom Register impl with custom Request DTO, logic and user model, I'd prefer to refactor the existing impl into reusable methods that they can pick and choose to use in their own Service as done in [this commit](https://github.com/ServiceStack/ServiceStack/commit/8ddae6dcf1582e0f0a2be2707066017268ef67f6) where reusable methods they may want to use is moved to a common `RegisterServiceBase` class.

We only need to support the popular use-case of creating new users since using the Register Service to update an existing user is uncommon and disabled by default.

So now to implement a custom registration service that validates and saves a new `NetworkName` property in their custom `AppUser` data model we can provide a starting pattern that makes use of existing functionality which they are free to customize and modify as they see fit, e.g:

```csharp
public class AppUser : UserAuth
{
    public string NetworkName { get; set; }
}

[Route("/register")]
public class RegisterUser : Register
{
    public string NetworkName { get; set; }
}

public class RegisterUserService : RegisterServiceBase
{
    public async Task<object> PostAsync(RegisterUser request)
    {
        if (string.IsNullOrEmpty(request.NetworkName))
            throw new ArgumentNullException(nameof(request.NetworkName));
        
        var session = await GetSessionAsync();
        if (await UserExistsAsync(session))
            throw new NotSupportedException("You're already registered");

        var newUser = (AppUser)ToUserAuth(request);
        newUser.NetworkName = request.NetworkName;

        await ValidateAndThrowAsync(request);
        var user = await AuthRepositoryAsync.CreateUserAuthAsync(newUser, request.Password);
        await RegisterNewUserAsync(session, user);
        
        var response = await CreateRegisterResponse(session, 
            request.UserName ?? request.Email, request.Password, request.AutoLogin);
        return response;
    }
}
```

We also don't need a plugin, allowing them to create a standard ServiceStack Service would be easier to reason about.

This change is now on MyGet so it's available to the existing project templates which are pre-configured to use the pre-release packages.

PR #1251, opened 2021-09-12
Fix ReDoS

ready-research:
Fix ReDoS

Reported in https://www.huntr.dev/bounties/dc853455-ecc7-47c9-9b82-ca7ab7970dc5/  Access this using your GitHub on the top right corner of huntr page. If you are able to access then you can just click on `mark as valid` and also you can `confirm the fix`.


mythz:
These assets are copied verbatim from swagger-ui which has not been updated:
https://github.com/swagger-api/swagger-ui/blob/2.x/lib/marked.js

Any changes need to happen upstream to swagger-ui project before we can copy the changes.

PR #1254, opened 2021-11-10
#767 Add support for custom select in AutoQuery filters.

AVee:
Note that I wasn't able to run all tests (no SQL Server on hand and only .NET core), but those that do run succeed.


mythz:
Brilliant, feature much appreciated, thx 👍

PR #1255, opened 2021-11-12
Handle removing Redis subs separately to NotifyOnLeave 

Layoric:
Handle removing Redis subs separately to NotifyOnLeave since that event can be skipped via feature flag.

Still need to add tests that confirm redis keys for subscriptions still being cleaned up regardless of option.


Layoric:
Added a test to show subscriptions stored in Redis now getting removed regardless of the `NotifyChannelOfSubscriptions = false` option, since there is now a new `OnRemoveSubscriptionAsync` event with a default to cleanup separately to the OnLeave notification. 

This is to resolve endless growth of `sse:ids:*` keys that if combined with use of methods like `GetAllSubscriptionInfos` will result in large number of Redis commands for expired sessions.

src/ServiceStack/ServerEventsFeature.cs
```diff
@@ -1324,6 +1326,9 @@ async Task DoAsyncTasks(CancellationToken token = default)
 
                     if (NotifyChannelOfSubscriptions && sub.Channels != null && NotifyLeaveAsync != null)
                         await NotifyLeaveAsync(sub).ConfigAwait();
+                    
+                    if (OnRemoveSubscriptionAsync != null)
+                        OnRemoveSubscriptionAsync(sub).ConfigAwait();
```
mythz, line 1331: need to `await` here otherwise PR LGTM 👍

Layoric: Ahh good catch, added.

PR #1260, opened 2022-01-06
Update HttpRequestExtensions.cs

glaidler:
Hi - we use BasicMessage for a variety of things, and this throws NRE if the AbsoluteUri is null - just an idea, please reject if this will break other bits.


mythz:
LGTM! thx for the PR 👍

mythz:
FYI this is now available from the latest [v5.13.3+ on MyGet](https://docs.servicestack.net/myget) if needed.

PR #1261, opened 2022-01-11
Add metadata for required fields to flow down to AutoForm.html.

Layoric:
Small change to support autoform to pickup properties attributed as required.


mythz:
LGTM, thx

PR #1273, opened 2022-04-05
Use dialect to format sql in DoesTableExist and DoesColumnExist funct…

aiyyo:
…ions


mythz:
Thx for the PR 👍 

If needed, this change is now available from the latest v6.0.3 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1274, opened 2022-04-28
pass through db parameter to update internal

kustnodi:
The PartialUpdate AutoCrud functions didn't pass through the db parameter to UpdateInternal


mythz:
Thanks for the PR 👍 

This change is now available from v6.0.3 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1278, opened 2022-05-23
Avoid NullReferenceException from property ServiceStackController.Redis

davecanderson:
When `IRedisClientsManager` hasn't been registered accessing `ServiceStackController.Redis` requires a try/catch block.

Added null conditional to `HostContext.AppHost.GetRedisClient()` to avoid the NullReferenceException. Method already used `TryResolve<IRedisClientsManager>()` but had no check for the result. Will allow a simple null check on property instead of try/catch block.


mythz:
Since this PR isn't mergeable due to conflicts, I've added it in this commit:

https://github.com/ServiceStack/ServiceStack/commit/1e4cc1ee4ba8b800d41a6ebc98c806e4b5627c07

I'd recommend doing a clean checkout to submit clean PR's in future.

This change is now available from [v6.1.1 on MyGet](https://docs.servicestack.net/myget)

PR #1288, opened 2022-07-21
Fix locode check for when auth is required but not specific role

Layoric:
Fix locode check for when auth is required but not specific role or permission is. Related to [issue raised in the customer forums](https://forums.servicestack.net/t/authentication-locode-and-roles/10839).


mythz (approved):
IMO guard should be before the test, but results in same behavior

PR #1291, opened 2022-08-08
prevent bucket re-creation error

rmt2021:
As the AWS bucket re-creation may not happen immediately, I add a thread sleep for 5 seconds after the bucket is deleted and before its creation. AWS doc:
https://aws.amazon.com/premiumsupport/knowledge-center/s3-conflicting-conditional-operation/

What's more, the `us-east-1`region will behave differently with others due to some legacy issues, so I think it is better to use other regions (us-east-2) instead of the us-east-1, as documented:
https://github.com/aws/aws-sdk-net/issues/2083


mythz:
Thx for the PR & background info! 👍

PR #1297, opened 2022-11-03
Fix redis connection string parsing bug

jklemmack:
Azure Redis Cache automatically generates passwords based on Base64 conversion, which add an "=" to the end of the password string.  This is not properly handled by `RedisExtensions.ToRedisEndpoint()`.

### DIFF
--- a/ServiceStack.Redis/src/ServiceStack.Redis/RedisExtensions.cs
+++ b/ServiceStack.Redis/src/ServiceStack.Redis/RedisExtensions.cs
@@ -62,7 +62,7 @@ namespace ServiceStack.Redis
                 var qsParams = qsParts[1].Split('&');
                 foreach (var param in qsParams)
                 {
-                    var entry = param.Split('=');
+                    var entry = param.SplitOnFirst('=');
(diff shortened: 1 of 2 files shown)


mythz:
Fix is fine but shouldn't be needed as values on the query string should be URL encoded which can easily be done with the Set/AddQueryParam ext methods, e.g:

```csharp
var host = "host.com:6123".AddQueryParam("password","U2VydmljZVN0YWNrIQ==");
```

mythz:
FYI this change is now available from v6.4.1+ that's now [available on MyGet](https://docs.servicestack.net/myget).

thx for the PR!

jklemmack:
> Fix is fine but shouldn't be needed as values on the query string should be URL encoded which can easily be done with the Set/AddQueryParam ext methods, e.g:
> 
> ```cs
> var host = "host.com:6123".AddQueryParam("password","U2VydmljZVN0YWNrIQ==");
> ```

Query string is stored in secrets and I can't reliably access it.  A certain other popular Redis library from a popular Q&A site has failed us and I am evaluating ServiceStack as a near drop-in replacement.

PR #1299, opened 2022-11-08
Support Sql.In with the constant first argument

OlegNadymov:


mythz:
thx for the PR, this change is now available from v6.4.1+ that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1304, opened 2022-12-05
Custom byte serializer

trejjam:
Hi, I hit some issues with ServiceStack.Text serializer and I would like to control the process. Now the `ToBytes` and `ToMessage` are extension methods making them difficult to customize. I extrapolate them to the new `ServiceStackTextMessageByteSerializer`. I believe it's without BC breaks because all constructor signatures were left untouched, I only added new ones.

Although it makes sense to add ServiceStackTextMessageByteSerializer to DI to prevent creating multiple instances, constructors without serializer can be made obsolete with a future main release.

### DIFF
--- a/ServiceStack/src/ServiceStack.Client/MessageExtensions.cs
+++ b/ServiceStack/src/ServiceStack.Client/MessageExtensions.cs
@@ -1,74 +1,9 @@
-using System;
-using System.Collections.Generic;
-using System.Threading;
 using ServiceStack.Messaging;
-using ServiceStack.Text;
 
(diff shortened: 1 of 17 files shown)


mythz:
Unfortunately this is too destructive and creates way too much boilerplate for such a small feature.

Instead I've modified all call sites to use an overridable singleton in this commit:
https://github.com/ServiceStack/ServiceStack/commit/816939931f4b0f79805c6ef269517b526dcb59d0

trejjam:
Understood, thank you!

PR #1305, opened 2022-12-29
ViewPage implementation to support @page directive in ASP.Net Razor Pages.

dandcg:
ViewPage currently inherits from RazorPage, which supports cshtml in general and not Razor Pages.

Ref: [https://github.com/aspnet/Mvc/issues/6769 1](https://github.com/aspnet/Mvc/issues/6769)

I’ve created a version of ViewPage (RazorViewPage) that derives from Microsoft.AspNetCore.Mvc.RazorPages.Page instead, and this only required a few tweaks to get working.

Now the ‘The name ‘PageContext’ does not exist in the current context’ error is gone.


dandcg:
Current a copy of the existing derived class with modifications - would utilising default implementations within an interface across both different classes be more maintainable?

mythz:
great thx looks like we're going with the alt base class approach, I didn't want to merge the other commits in this PR so I just added the base class in [this commit](https://github.com/ServiceStack/ServiceStack/commit/59a28085c578d439736f45dc33b60308bea40790).

I've called it `RazorPage` to be closer to "Razor Pages" and had to rename Request/Response to avoid conflicts with base properties.

This change is available from v6.5.1+ that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #1312, opened 2023-03-06
fix invalid 2d array type declaration in TypeScriptGenerator

migajek:
We found out the TypeScriptGenerator doesn't play well with 2d+ arrays: 

```c#
var gen = new TypeScriptGenerator(null);
var result = gen.Type("String[][]", null);
```

result is `string[]` instead of `string[][]`


mythz:
Cheers 👍

mythz:
If needed this change is now available from the latest v6.6.1+ that's now [available on MyGet](https://docs.servicestack.net/myget)

thx for the PR 👍

PR #1314, opened 2023-03-31
Make EscapeWildcards as opt-out

pedoc:
## Motivation:
In some cases we don't want to execute the EscapeWildcards operation, but in order to be consistent with the previous behavior, we implement it as an opt-out.


mythz:
Cheers 👍

PR #1320, opened 2023-05-09
Pass connection parameter to called UpdateInternalAsync

kustnodi:
Encountered problem when trying to do a bulk soft delete in a loop.  There was an exception thrown in the middle of the update but when I rolled back the transaction, the updates for the soft deletes up to that point were committed.

The delete method delegates to the UpdateInternal method but did not pass its db parameter along causing the existing connection to reset and transaction is unable to rollback the updates.


mythz:
Thanks for the PR 👍

mythz:
This is now available from v6.8.1+ that's now [available on MyGet](https://docs.servicestack.net/myget)

kustnodi:
Thanks!

PR #1321, opened 2023-05-12
Do not skip tags for SqlExpressions that have Limit

canro91:
OrmLite generates an invalid SQL query when tagging a `SqlExpression` that has a Limit, with `SqlServerOrmLiteDialectProvider`.

For example, the expression `db.From<SimpleModel>().Limit(1).TagWith(AnAwesomeQuery)` generates 

```sql
SELECT TOP 1wesomeQuery

SELECT "Id", "Foo", "Bar" 
FROM "SimpleModel"
```

which throws "_System.Data.SqlClient.SqlException : Invalid column name 'wesomeQuery'._"


mythz:
Thx for the PR & Tests! 👍 

This change is available from v6.8.1+ that's now [available on MyGet](https://docs.servicestack.net/myget)

PR #1326, opened 2023-08-02
Fix first element in list losing leading whitespace when deserialized

KuzonCode:
Found an issue where deserializing a list that had leading whitespace would lose the leading whitespace only on the first element.


mythz:
Thx for the PR! This change is now available from **v6.10.1* that's now available in the [Pre-Release NuGet Packages](https://docs.servicestack.net/pre-release).

PR #1332, opened 2023-11-23
Funq: Try resolving service from Adapter first, before applying Lazy Resolver

migajek:
MS DI doesn't have the named services feature (yet) therefore it's quite common pattern to overcome it with factory func:

```
.AddSingleton<Func<string, Func<IService>>>(services => name => name switch {
        "foo" => () => services.Resolve<FooService>(),
        "bar" => () => services.Resolve<BarService>()
        })
```

however the injection of such factory function into ServiceStack services failed mysteriously.

This is because Funq decided to wrap it in Lazy Resolver before checking the presence of the exact registration in an Adapter.

### DIFF
--- a/ServiceStack/src/ServiceStack/Funq/Container.cs
+++ b/ServiceStack/src/ServiceStack/Funq/Container.cs
@@ -428,10 +428,32 @@ namespace Funq
                 container = container.parent;
             }
 
-            if (entry == null)
+            if (entry != null)
             {
-                var genericDef = typeof(TService).FirstGenericTypeDefinition();
-                if (genericDef != null && genericDef.Name.StartsWith("Func`")) //Lazy Dependencies
+                if (entry.Reuse == ReuseScope.Container && entry.Container != this)
+                    entry = SetServiceEntry(key, ((ServiceEntry<TService, TFunc>)entry).CloneFor(this));
+            }
+            else
+            {
+
+                if (TryResolveFromAdapter(out var instance))
+                {
+                    try
+                    {
+                        return new ServiceEntry<TService, TFunc>(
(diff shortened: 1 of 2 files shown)


mythz:
This PR undoes a lot of the `.csproj` changes we've done to upgrade them to .NET 8, can you apply your changes from a clean checkout so that it contains only the code changes from the latest code base please.

migajek:
Jeez sorry my bad. It was supposed to be local only. Will fix it asap

W dniu czw., 23.11.2023 o 12:52 Demis Bellot ***@***.***>
napisał(a):

migajek:
Updated. There don't seem to be checks / CI at all, though

mythz:
This broke a lot of tests, e.g. in [IocServiceTests.cs](https://github.com/ServiceStack/ServiceStack/blob/main/ServiceStack/tests/ServiceStack.WebHost.Endpoints.Tests/IocServiceTests.cs) so I've had to replace the approach in [this commit](https://github.com/ServiceStack/ServiceStack/commit/cc151ce3e5e55a821a6999ba0d7a0f8af08db215) to just always checking the adapter for any `Func<>` resolutions first.

Since Func uses `Func<>` for its lazy resolutions, the test for `Resolve<Func<T>>()`  couldn't be retained since if there exists an Adapter it will always return a lazy implementation that resolves the resolution from the Adapter. So I've just changed to executing the resolved lazy resolution:

```csharp
Assert.Catch(() => container.Resolve<Func<string>>()());
```

Note this is an unreliable behavior to rely on in Funq since it represents a Lazy check for `String` not eager check for `Func<String>`.

PR #1333, opened 2023-12-11
Can use constant on left

OlegNadymov:
I added the test ``Can_use_constant_as_param_on_left`` where I show my not working case.
Is it possible to add this possibility: transform constant on the left into Sql params, not into string db values?
In the case of datetime type in MS SQL, it doesn't work with the exception:
```
Conversion failed when converting date and/or time from character string
```


mythz:
This PR was too dirty to accept (needs a clean checkout so PR only contains changes), but I've added support for this test in [this](https://github.com/ServiceStack/ServiceStack/commit/e85caeab5d636480a53df86182bfc004d336fdb0) and [this commit](https://github.com/ServiceStack/ServiceStack/commit/f21e73a2de93681bb610dc4d15e3e4e4aaf7a1ba).

This change is available from v8.0.1+ that's now available in [pre release packages](https://docs.servicestack.net/pre-release).

OlegNadymov:
My bad! I did the PR in a hurry. If I need to do another PR, I'll do it cleaner.
@mythz, thank you for the fix! 👍

PR #1341, opened 2024-02-15
Update AutoQueryGrid.razor.cs

sbosell:
Provide a callback for AutoQueryGrid when the data is returned from the API so that one can to do calculations or show something external to the grid based on the data.


mythz:
Thx for the PR! If needed this is available from v8.13+ that's now available in our [Pre Release packages](https://docs.servicestack.net/pre-release).

PR #1342, opened 2024-02-28
Update path utils to safeguard whitespaces.

tungtle001:
Updated path combining to safeguard from whitespaces in addition to the existing null and empty checks. This is so it doesn't combine unnecessary whitespaces and forward slashes to the path.


tungtle001:
Contributor license agreement submitted. Thanks!

mythz:
Cheers!

PR #1343, opened 2024-03-10
Add missing ServiceStack.OrmLite.Oracle.Core.csproj

pedoc:
1. Other libraries such as ``Mysql``, ``MysqlConnector``, etc. all have ``*.Core`` versions.

2. There is currently a ``GetLastSql`` extension function, but this is not intuitive for sql with parameters, and it can even be said to be meaningless, because even if the sql is obtained from the log, no verification and debugging can be performed at the database level.
Therefore, relevant infrastructure has been added so that complete sql statements can be obtained


mythz:
Thanks for the PR. We didn't have Oracle.Core because it's a community provided provider which we don't support ourselves, but if you're willing to maintain it for your own purposes we'll happily include it.

This is now available from v8.2.3+ which is now available in the [Pre Release Packages](https://docs.servicestack.net/pre-release)

PR #1348, opened 2024-06-14
Allow to deserialize unknown sub type

Tarboeuf:
During my update from ServiceStack 5.9.2 to 8.2.2 I figured out that the behavior changed.
The test ArrayThatContainsUnknownType_ShouldContainNullItem that's failing without the fix explains my issue. (       Assert.That(stub.Stubs, Is.Not.Null);)

To be coherent with DeserializeTypeRefJson., You can check the nullability of the type before calling AssertAllowedRuntimeType which fails with an exception


mythz:
Thx for the PR 👍 

If needed this change is now available from our [Pre Release Packages](https://docs.servicestack.net/pre-release).

PR #1350, opened 2024-09-25
Support evaluate InvocationExpression like MethodCallExpression

pedoc:
I accidentally found that ormlite generated incorrect sql for ``Func<T>`` related calls. This modification solved the problem, but I am not sure if there are any omissions or hidden dangers. Please check it out. 

Thank you


mythz:
Cheers, thx for the PR!

PR #1353, opened 2024-11-26
using >=6.0.0 for net6.0 reference (instead of 8.0.0).

pauldotknopf:
using >=6.0.0 for net6.0 reference (instead of 8.0.0)

the azure functions runtime doesn't handle 8.0.0 references in a 6.0 runtime. Not sure why, but it doesn't.

### DIFF
--- a/ServiceStack.Text/src/Directory.Build.props
+++ b/ServiceStack.Text/src/Directory.Build.props
@@ -56,4 +56,17 @@
     <DefineConstants>DEBUG</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net472' ">
+    <MicrosoftFamilyLibraryVersions>8.0.0</MicrosoftFamilyLibraryVersions>
+  </PropertyGroup>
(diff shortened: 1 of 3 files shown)


pauldotknopf:
This was the exact error I was getting, running Azure Functions runtime 4.0 with net6.0:

```
Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified
```

mythz:
thx for the PR, we'll test it out after tonight's release.

mythz:
This issue should only be with DiagnosticSource since `Microsoft.Bcl.AsyncInterfaces` is only used in **netstandard2.0** builds which isn't referenced in .net6.0 projects so it wouldn't impact your .net6.0 App. 

Also any solution we adopt should apply to the entire solution and all projects use framework conditionals to control which dependencies they reference so I've instead made the change for DiagnosticSource in this commit:
https://github.com/ServiceStack/ServiceStack/commit/d34f78b68a22db701679a6ab5094f2b845a9a6b1

If needed this is now available in the latest v8.5.1 [pre-release packages](https://docs.servicestack.net/pre-release)

PR #1354, opened 2024-11-28
Coalesce with nullable bool

OlegNadymov:
MS Sql doesn't support the statements like this:

```sql
SELECT * FROM _mytable
WHERE not COALESCE(_is_bool, 0)
```

It has to be:
```sql
SELECT * FROM _mytable
WHERE COALESCE(_is_bool, 0) = 0
```

There is unit test in this PR: 
`Can_use_coalesce_with_nullable_bool`


mythz:
Cheers :+1:

OlegNadymov:
@mythz thank you!
I'd like to know if you're going to fix this "Coalesce" problem.

OlegNadymov:
I think the same problem can be in PostgreSQL.
I've checked in https://pgplayground.com/

mythz:
The test passes in PostgreSQL

OlegNadymov:
Great! 👍

PR #1355, opened 2024-12-12
Set GatewayRequest.Files from IRequest in FromRequest

DeonHeyns:
Currently when making a call through the Gateway the Files from the Request object is not set on the GatewayRequest


mythz:
Cheers :+1:  If needed this fix is now available from v8.5.3+ that's now available in [pre release packages](https://docs.servicestack.net/pre-release)

PR #1356, opened 2025-04-08
support getting connection strings with unsupported IDbConnectionFact…

pauldotknopf:
…ory types


mythz:
thx for the PR!

If needed this change is now available in the [pre-release packages](https://docs.servicestack.net/pre-release)

PR #1357, opened 2025-05-02
Pass inner exception to exceptions thrown in JsonApiClient

pascalberger:
When during `JsonApiClient.SendAsync` an exception was thrown, it was catched and instead an `WebServiceException` thrown. The original exception was ignored, same for the `HttpResponseMessage` which was passed to the internal `CreateException` method.

We had cases where the request itself was successful with a HTTP 200, but an `WebServiceException` was thrown. What we saw in (OpenTelemetry) logs:

exception.message:
```
Exception of type 'ServiceStack.WebServiceException' was thrown.
```

exception.stacktrace:
```
500 Code: Exception of type 'ServiceStack.WebServiceException' was thrown., Message: Exception of type 'ServiceStack.WebServiceException' was thrown.
```

This PR passes the actual exception as inner exception to the created `WebServiceException` and also sets the message to the message of the inner exception, so that there is an exception in case `ResponseStatus.Message` is not available.

### DIFF
--- a/ServiceStack/src/ServiceStack.Client/JsonApiClient.cs
+++ b/ServiceStack/src/ServiceStack.Client/JsonApiClient.cs
@@ -803,7 +803,7 @@ public class JsonApiClient : IJsonServiceClient, IHasCookieContainer, IServiceCl
         }
         catch (Exception e)
         {
-            throw CreateException(httpRes, e);
+            throw CreateException(e);
         }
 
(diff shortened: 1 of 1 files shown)


mythz:
thx for the PR!

This change is available from v8.7.3+ which is now available in the [pre release packages](https://docs.servicestack.net/pre-release)

PR #1358, opened 2025-05-06
If path is rooted ignore Content Root Path

DeonHeyns:
Fixes an issue where if the DbDir is a root path that the monthly database doesn't get incorrectly prefixed with the content root path of the app


mythz:
thx for the PR, if needed it's now available from v8.7.3+ that's now available in our [pre-release packages](https://docs.servicestack.net/pre-release)

PR #1361, opened 2025-06-24
feat: add kingbase support

pedoc:
PS:
A commercial database based on postgre and supporting mysql oracle sqlserver postgre dialects

more info:
https://kingbase.com.cn/


pedoc:
> Unfortunately since I don't understand Chinese and don't know anything about this provider I don't want to add it to the primary code base which I wont know how to support so it'd be better if the provider was maintained in a separate repo that you own and have full control over.
> 
> I'm happy to accept any changes to OrmLte's code base that you'd need to add support this provider from an external repo.

Understood😊

PR #1362, opened 2025-06-26
feat: Added a parameterless RowCount read API and fixed some tests.

pedoc:


mythz:
Instead of changing existing method to `ToRowCountStatement(string innerSql, bool innerSqlAsSubQuery=true)` can you add an extra method like `ToRowCountInStatement(innerSql)`

pedoc:
> Instead of changing existing method to `ToRowCountStatement(string innerSql, bool innerSqlAsSubQuery=true)` can you add an extra method like `ToRowCountInStatement(innerSql)`

The current changes have destructive changes to the public API (new parameters are added), but they are enough to solve the current problem

It may be more appropriate to put a new PR to add a new API.

mythz:
Hi I've added a new RowCount API for discoverability in this commit https://github.com/ServiceStack/ServiceStack/commits, which didn't require changing the DialectProvider as you could just use an SQL Expression, e.g:

```csharp
public static long RowCount<T>(this IDbConnection dbConn)
{
    return dbConn.SqlScalar<long>(dbConn.From<T>().Select(Sql.Count("*")));
}
```

But this isn't needed as it's functionally existing equivalent to the existing `Count`, e.g:

```csharp
var rowCount = db.Count<LetterFrequency>();
```

This change is available in the latest v8.8.1+ that's now available in the [pre-release packages](https://docs.servicestack.net/pre-release).

PR #1366, opened 2025-08-14
Make PrepareWebRequest virtual

SaarCohen:
That way we can renew the token if expired in our own service client, which derives from this one


mythz:
Thx for the PR, if needed this change is now available from the [pre-release packages](https://docs.servicestack.net/pre-release)

PR #1367, opened 2025-08-18
When the actual value on duty is a string NULL, the SQL expression generated should not be "is not null".

vla:


vla:
@mythz 

Oh! Sorry, I didn't pay attention to the pass rate of the unit tests. Here, I'll provide some code to explain the issue.

```code
await conn.UpdateAsync<car_device>(new
 {
     no = "xxx",
 }, q => q.types != "NULL");
```
This code will generate "where types is not null", but in fact, what is needed is "where types <> 'NULL'".

mythz:
Should be resolved in this commit https://github.com/ServiceStack/ServiceStack/commit/0a224914f273c0638e09035d4305a0452f099c9f

This change is available from the latest v8.8.1+

PR #1377, opened 2026-07-21
Fix Firebird OrmLite type converters + add Firebird 5 dialect

madaleno:
## Summary
Fixes a set of Firebird OrmLite type converters that emit invalid column types or round-trip incorrectly against a real Firebird server, and adds a canonical Firebird 5 dialect.

## Bugs fixed (ServiceStack.OrmLite.Firebird)
| .NET type | before | after |
|-----------|--------|-------|
| `double` | `FLOAT` (32-bit, precision loss) | `DOUBLE PRECISION` |
| `bool` | `RemoveConverter<bool>` / `INTEGER` bind (InvalidCast) | native `BOOLEAN` (FB3+) |
| `Guid` | `CHAR(16) OCTETS`, asymmetric read swap | symmetric round-trip |
| `DateTime` | `LOCALTIME` (context var, not a column type) | `TIMESTAMP` |
| `DateOnly` | `DATETIME` (not a Firebird type) | `DATE` (net6+) |
| `TimeOnly` | `BIGINT` ticks (not a `TIME` column) | `TIME` (net6+) |

## Firebird 5 dialect
Adds `Firebird5OrmLiteDialectProvider` + `Firebird5Dialect`. FB5 (ODS 13.1) adds no new **reserved** words over FB4 (only non-reserved keywords), so it inherits `Firebird4` unchanged as the canonical FB5 dialect / extension point.

## Tests
Adds `FirebirdTypeConverterTests`: round-trips every affected type (plus false-bool / empty-Guid), asserts generated DDL column types, and verifies the FB5 dialect inherits the converters. Verified **4/4 green against a real Firebird 5.0.2 server on the stock `FirebirdSql.Data.FirebirdClient`**. `DateOnly`/`TimeOnly` are `#if NET6_0_OR_GREATER` guarded.

## Notes
- No change to the FB4 reserved-word list (already correct upstream).
- No new dependencies; converter fixes work on the stock client

### DIFF
--- a/ServiceStack.OrmLite/src/ServiceStack.OrmLite.Firebird/Converters/FirebirdBoolConverter.cs
+++ b/ServiceStack.OrmLite/src/ServiceStack.OrmLite.Firebird/Converters/FirebirdBoolConverter.cs
@@ -1,13 +1,31 @@
 ﻿using System;
+using System.Data;
 using ServiceStack.OrmLite.Converters;
 
 namespace ServiceStack.OrmLite.Firebird.Converters
 {
+    // Firebird < 3 has no BOOLEAN type -> store bool as INTEGER (base/legacy dialect).
     public class FirebirdBoolConverter : BoolAsIntConverter
     {
         public override string ColumnDefinition
         {
             get { return "INTEGER"; }
         }
+    }
+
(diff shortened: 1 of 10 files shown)


mythz:
Appreciated!

PR #1379, opened 2026-09-08
Convert BigInteger in ConvertNumber (Firebird 4+ INT128)

madaleno:
ConvertNumber reaches Convert.ToInt64/ToInt32/... for every numeric target, and those throw on a BigInteger rather than converting it:

    InvalidCastException: Unable to cast object of type 'System.Numerics.BigInteger'
                          to type 'System.IConvertible'

BigInteger does not implement IConvertible, by design.

Providers do return it. Firebird 4 introduced INT128 and widens SUM() of a BIGINT to it, and the stock FirebirdClient correctly surfaces INT128 as BigInteger. Measured on Firebird 5.0.4:

    SUM(SMALLINT)/SUM(INTEGER)  -> Int64        SUM(BIGINT)             -> BigInteger
    SUM(NUMERIC(18,0))          -> Decimal      SUM(OCTET_LENGTH(blob)) -> BigInteger
    COUNT(*)                    -> Int64

So COUNT works and summing a BIGINT throws, at runtime, in whichever query hits it first. OCTET_LENGTH of a BLOB is BIGINT, so summing blob sizes reproduces it without anything in the schema being declared INT128 - which is how it was found, on a query totalling image sizes.

BigInteger is now converted explicitly, ahead of the existing switch. The casts are checked, so a value genuinely too large for the target still throws OverflowException rather than silently truncating.

Tests:
- ConverterTests (no database, runs everywhere): conversion to each numeric target, negative and zero, and an overflow assertion for a value that does not fit.
- FirebirdInt128SumTests (Firebird 4+): SUM of a BIGINT column read as long/int/decimal/double, SUM of OCTET_LENGTH over a BLOB, and SUM over no rows. A BIGINT column is required - an INTEGER one does not widen to INT128 and would pass with or without this change.

Verified against a real Firebird 5.0.4 server on the stock NuGet FirebirdClient: 4/4 pass with the fix; reverting just the converter change fails 3 of the 4 with the exception above (the fourth sums no rows, so it returns NULL before any conversion).

### DIFF
--- a/ServiceStack.OrmLite/src/ServiceStack.OrmLite/IOrmLiteConverter.cs
+++ b/ServiceStack.OrmLite/src/ServiceStack.OrmLite/IOrmLiteConverter.cs
@@ -123,6 +123,51 @@ public static class OrmLiteConverterExtensions
             return value;
 
         var typeCode = toIntegerType.GetUnderlyingTypeCode();
+
+        // BigInteger does NOT implement IConvertible, so every Convert.ToX below throws
+        // InvalidCastException ("Unable to cast object of type 'System.Numerics.BigInteger' to type
+        // 'System.IConvertible'") instead of converting it.
+        //
+        // Providers do hand it back. Firebird 4+ has INT128 and widens SUM() of a BIGINT to it, which the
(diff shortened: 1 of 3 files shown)


mythz:
Great, thx!

PR #1382, opened 2026-09-11
Fix string[] Contains translated to LIKE under C# 14 and interpret one-shot Evaluate expressions

albyrock87:
Follow-up to ServiceStack/Issues#828. While bisecting that regression to 10a2ecf41 we found a second regression from the same commit in OrmLite, plus a cheap way to recover most of the performance lost when the slow-compile cache was removed in 9519cbb20.

## 1. `string[]` `Contains` is translated to `LIKE` under C# 14 (regression from 10a2ecf41)

With C# 14 (the default for `net10.0`) `array.Contains(x.Col)` inside an expression tree binds to `MemoryExtensions.Contains<T>(ReadOnlySpan<T>, T)` instead of `Enumerable.Contains`. `IsSpanContainsOnArray` tried to tell substring searches apart from collection lookups by looking at the type of the *searched value*. For a `string[]` collection that value is the string column, so the call was rejected, fell through to the string-method path and silently produced:

```sql
WHERE @1 like @2      -- params: '', '', '%System.String[]%'
```

That query matches nothing and throws no error. It affects captured `string[]` variables, `new[] { "a", "b" }.Contains(x.Col)` literals and `string[]` fields read through `this`. `int[]`, `Guid[]`, `List<string>` and `IEnumerable<string>` were unaffected, which is why the existing int-array tests didn't catch it. The same code on 40e6bf521 (the commit before 10a2ecf41) generated a correct `IN (@0,@1)`.

Fix in `SqlExpression`:
- classify on the span's element type instead: only `ReadOnlySpan<char>`/`Span<char>` overloads are substring searches, everything else is a collection lookup;
- unwrap the compiler-inserted array → span conversion (`Convert` node or `op_Implicit` call) before evaluating the collection, so the array itself goes through the cached expression compiler fast paths instead of a boxed ref struct;
- stop visiting the span argument, whose result was discarded.

Tests were committed first (they fail on `main` with the output above), then the fix.

## 2. Interpret one-shot `Evaluate` expressions instead of JIT compiling them

Since 9519cbb20 removed the slow-compile cache, every expression the fingerprint path can't handle (`NewArrayInit` for inline `Sql.In` params, `New`, `MemberInit`, `ListInit`) is compiled on each `Evaluate` call. `Evaluate` invokes the delegate exactly once, so emitting and JIT compiling a `DynamicMethod` for it is wasted work.

This compiles with `preferInterpretation: true` for the `Evaluate` path only. The cached paths are untouched and the reusable `Compile<TModel, TValue>` API still returns JIT compiled delegates. `net472` has no interpreter overload and keeps the previous behaviour.

Measured on .NET 10 (each op includes building the expression tree):

| Shape reaching the slow path | JIT (9519cbb20) | interpreted |
|---|---|---|
| `new object[] { orgId, 0 }` (inline `Sql.In` params) | 61 µs | 3.8 µs |
| `new Entity { Count = orgId }` | 53 µs | 4.0 µs |
| `new DateTime(y, m, d)` | 109 µs | 5.9 µs |

Allocations roughly halve and no `DynamicMethod` is emitted per call.

## Verification

- `ServiceStack.Common.Tests` `CachedExpressionCompilerTests`: 17 tests pass on net8.0 and net10.0; `ServiceStack.Common` builds for all targets including net472.
- `ServiceStack.OrmLite.Tests` (Sqlite): 1096 passed / 27 failed on this branch vs 1091 / 27 on `main`. The 27 failures are identical on both sides (`Can_filter_*_to_insert_date`, `Can_create_table_with_DefaultValues`, `Can_use_ToUpdateStatement_to_generate_inline_SQL`, all `SQLite Error 1: near ",": syntax error`) and are unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ServiceStack.OrmLite/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/ServiceStack.OrmLite/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -2912,20 +2912,35 @@ namespace ServiceStack.OrmLite
             if (!isMemoryExtensions)
                 return false;
 
-            // Check the second argument type to determine if this is string/char Contains or collection Contains
-            var secondArgType = m.Arguments[1].Type;
-
-            // If the second argument is a string or ReadOnlySpan<char>, this is string Contains (should use LIKE)
-            if (secondArgType == typeof(string) || secondArgType == typeof(char))
+            // The element type of the span (first argument) decides what kind of Contains this is:
+            // - ReadOnlySpan<char>/Span<char> overloads are substring searches (should use LIKE)
+            // - any other element type is a collection lookup (should use IN)
+            // The searched value (second argument) must not be used for this decision, otherwise a
+            // string[] collection searched with a string column would be misclassified as a LIKE.
+            var spanType = m.Arguments[0].Type;
+            if (spanType.IsGenericType &&
+                spanType.GenericTypeArguments.Length == 1 &&
+                spanType.GenericTypeArguments[0] == typeof(char))
                 return false;
 
-            if (secondArgType.Name == "ReadOnlySpan`1" && secondArgType.GenericTypeArguments.Length > 0 &&
-                secondArgType.GenericTypeArguments[0] == typeof(char))
-                return false;
-
-            // Otherwise, this is a collection Contains (should use IN)
-            // The second argument is the value being searched for (e.g., x.Id)
             return true;
+        }
+
+        private static Expression UnwrapSpanConversion(Expression e)
+        {
+            // C# 14 inserts an implicit array -> ReadOnlySpan<T>/Span<T> conversion, either as a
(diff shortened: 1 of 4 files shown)


mythz:
Thanks for the PR with analysis + tests! 

Can you please [Approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so we can merge this, thx.

mythz:
This change is now available from the latest v10.2.1+ that's now available in the latest [pre-release packages](https://docs.servicestack.net/pre-release).

As you already have the pre-release packages installed you'll need to [clear your NuGet packages cache](https://docs.servicestack.net/pre-release#redownloading-pre-release-packages) to download the latest version:

```bash
$ dotnet nuget locals all --clear
```

